PR #4, opened 2020-07-27
Bring in structure from cthaeh

pipermerriam:
## What was wrong?

Stubs out basic CLI application.

## How was it fixed?

Using `async-service` for long running process model.

CLI entry point avoids triggering any imports outside of standard lib and what is needed to enter into `trio.run`.

No actual application logic implemented yet

#### Cute Animal Picture

chickens-soccer


pipermerriam:
TODO: pexpect based tests.

Issue #64, opened 2020-08-26
Implement NetworkAPI.ping


pipermerriam:
## What was wrong?

The `ClientAPI` exposes a `ping` method that performs a round trip.  However, it returns the raw `InboundMessage[PongMessage]` response which is not as helpful since it is wrapped.

## How can it be fixed?

Implement `NetworkAPI.ping` which unwraps the response, returning only the `PongMessage`.

Issue #63, opened 2020-08-26
NetworkAPI.find_nodes(...)


pipermerriam:
## What was wrong?

The `ClientAPI` exposes a `find_nodes` method that does the full request/response cycle.  This however is not very useful in the application layer because it returns a `Tuple[IncomingMessage[FoundNodesMessage]]` which takes extra work to extract the ENR records.

## How can it be fixed?

Implement `NetworkAPI.find_node` which returns `Tuple[ENR, ...]`, aggregating the response messages for convenience.

Issue #61, opened 2020-08-26
Implement routing table management in NetworkAPI


pipermerriam:
## What was wrong?

The `NetworkAPI` should maintain the routing table

## How can it be fixed?

- The NetworkAPI should be initialized with some set of *bootnodes* which should be loaded into the routing table.
- A long running task should work to populate the routing table by periodically performing iterative lookups to random node ids.
- A long running task should monitor monitor the node ids in the routing table, periodically sending a ping to nodes to verify they are still reachable.  Unresponsive nodes should be removed from the routing table.

Issue #60, opened 2020-08-26
NetworkAPI.recursive_find_nodes(...)


pipermerriam:
## What was wrong?

Need to implement the recursive node finding algorithm.

## How can it be fixed?

Use previous implementation as baseline

https://github.com/ethereum/alexandria/blob/adba4114fbd5f707181da602abd977e008e463c9/alexandria/network.py#L99-L134

And the discovery v5 spec:

https://github.com/ethereum/devp2p/blob/0675791e69ef95ca941a426f50a34d0cfe5c54b4/discv5/discv5-theory.md#lookup-protocol

PR #66, opened 2020-08-26
Routing table management

pipermerriam:
## What was wrong?

The network API holds the routing table but currently does nothing to manage it.

## How was it fixed?

The network API now manages the routing table.

- upon booting it establishes connections with the provided bootnodes and bootsraps the routing table.
- a long running process periodically searches the network for a random node-id, bonding with the discovered nodes and adding them to the routing table.
- supporting functionality is implemented for both singular `find_nodes` requests and `ping` requests as well as recursive lookups.

#### Cute Animal Picture

preview16

### DIFF
--- a/ddht/kademlia.py
+++ b/ddht/kademlia.py
@@ -3,9 +3,21 @@ import functools
 import ipaddress
 import itertools
 import logging
+import math
 import random
 import struct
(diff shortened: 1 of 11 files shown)


pipermerriam:
Planning to break this up into smaller units.  This fixes all of #60 #63 #64 #61

PR #73, opened 2020-08-28
Add file logging, more control over what is logged

### LINKED ISSUE #14: Setup file logging for DEBUG level
## What was wrong?

DEBUG level logs are useful but noisy.

## How can it be fixed?

Do what we do in trinity and automatically log to a file at DEBUG level.

lithp:
Fixes #14.

I took some liberties here, let me know if you want to back any of them out:

- `-l` and `--level` are gone, and have been replaced with the more specific `--log-level-file` and `--log-level-stderr`. Those default to DEBUG and INFO, respectively.
- I changed `shortname` to `name`, to make it easier to figure out exactly which logger is spamming you, so it can be silenced with `LOGLEVEL`.
- I added a `LOGLEVEL` variable, which is inspired by [`PYTHONWARNINGS`](https://docs.python.org/3/library/warnings.html#describing-warning-filters), and gives precise control over which loggers are used. I begun by trying to implement it with `--level` and argparse, as Trinity does, but realized that the env variable would require a lot less code, and makes it easier to manage, using either an `.env` file or an `export`.

Here's a command which gets rid of all the upnp spam:

> LOGLEVEL=INFO:ddht.upnp,INFO:urllib3.connectionpool,INFO:Service,INFO:Action,INFO:ssdp,INFO:Device ddht --log-level-stderr 10

It's unwieldy to write this out every time. But after an `export` we can get all the benefit of not being spammed by upnp, while still running just

> ddht --log-level-stderr 10

In the future it might make sense to read configuration from `pyproject.toml`, but for now this seems nice and convenient.

#### Cute Animal Picture

beaver

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,20 @@ Read more in the [documentation on ReadTheDocs](https://ddht.readthedocs.io/). [
 pip install ddht
 ```
 
+To run it:
+
+```sh
+ddht
+```
+
+`--help` will tell you about the arguments `ddht` accepts. The LOGLEVEL environment
+variable can be used to control which log messages are emitted. For example, to suppress
+unimportant messages from the Packer you can run:
+
+```sh
+LOGLEVEL=WARNING:ddht.v5.packer.Packer ddht
+```
+
 ## Developer Setup
 
 If you would like to hack on ddht, please check out the [Snake Charmers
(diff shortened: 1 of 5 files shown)


pipermerriam:
btw, `make lint-roll` is your friend.

lithp:
Added a note to the README and appeased lint.

pipermerriam:
Just realized this is dumping logs in the current working directory.

I think we need these to be in `$XDG_DATA_HOME/ddht/logs/`

pipermerriam:
Also, logs should rotate which is pretty easy to setup.

lithp:
Adding this in a sec!

PR #80, opened 2020-08-29
Use a RotatingFileHandler for logs, default to XDG_DATA_HOME

lithp:
Small fixes, missed in #73 

- Switch to using a rotating file handler.
- Default to throwing logs into `.local/share/ddht/logs`

#### Cute Animal Picture

smol-chipmunk

### DIFF
--- a/ddht/main.py
+++ b/ddht/main.py
@@ -39,7 +39,7 @@ async def main() -> None:
         logdir = log_file = get_xdg_ddht_root() / "logs"
         logdir.mkdir(parents=True, exist_ok=True)
 
-        log_file = logdir / "ddht.logfile"
+        log_file = logdir / "ddht.log"
     else:
         log_file = args.log_file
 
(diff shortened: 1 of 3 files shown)


ddht/main.py
```diff
@@ -39,6 +35,18 @@ async def main() -> None:
                 f"not under the $XDG_DATA_HOME: {boot_info.base_dir}"
             )
 
+    if args.log_file is None:
+        logdir = log_file = get_xdg_ddht_root() / "logs"
+        logdir.mkdir(parents=True, exist_ok=True)
+
+        log_file = logdir / "ddht.logfile"
```
pipermerriam, line 42: I think that `.log` is a more common extension (and it matches what we're doing in trinity)

pipermerriam, line 42:
```suggestion
        log_file = logdir / "ddht.log"
```

PR #90, opened 2020-09-08
Implement core JSON-RPC aPI

pipermerriam:
## What was wrong?

Need a JSON-RPC API for interfacing with the running application.

## How was it fixed?

Some copy/pasta with minor modifications from the alexandria codebase.

Implemented a web3.py module that can be used to expose high level API endpoints using the web3.py library.

Implemented endpoints for basic node info and routing table info.

#### Cute Animal Picture

stuckanimals4

### DIFF
--- a/ddht/rpc.py
+++ b/ddht/rpc.py
@@ -4,13 +4,14 @@ import io
 import json
 import logging
 import pathlib
-from typing import Any, Dict, Generic, Mapping, Optional, Tuple, TypeVar, cast
+from typing import Any, Dict, Generic, Mapping, Tuple, TypeVar, cast
 
 from async_service import Service
 from eth_utils import ValidationError
 import trio
 
 from ddht.abc import RPCHandlerAPI, RPCRequest, RPCResponse
+from ddht.exceptions import DecodingError
 from ddht.typing import JSON
 
 NEW_LINE = "\n"
@@ -28,6 +29,9 @@ logger = logging.getLogger("ddht.rpc")
 decoder = json.JSONDecoder()
 
 
+MAXIMUM_RPC_PAYLOAD_SIZE = 1024 * 1024  # 1 MB
+
+
 async def read_json(
     socket: trio.socket.SocketType,
     buffer: io.StringIO,
@@ -43,7 +47,15 @@ async def read_json(
         if bad_prefix:
             logger.info("Client started request with non json data: %r", bad_prefix)
             await write_error(socket, f"Cannot parse json: {bad_prefix}")
-            continue
+            raise DecodingError(f"Invalid JSON payload: prefix={bad_prefix}")
+
+        if len(raw_request) > MAXIMUM_RPC_PAYLOAD_SIZE:
+            error_msg = (
+                f"RPC payload exceeds maximum size: {len(raw_request)} "
+                f"> {MAXIMUM_RPC_PAYLOAD_SIZE}"
+            )
+            await write_error(socket, error_msg)
+            raise DecodingError(error_msg)
 
         try:
             request, offset = decoder.raw_decode(raw_request)
@@ -57,9 +69,6 @@ async def read_json(
                 await trio.sleep(0.01)
             continue
 
-        # TODO: more efficient algorithm can be used here by
-        # manipulating the buffer such that we can seek back to the
-        # correct position for *new* data to come in.
         buffer.seek(0)
         buffer.write(raw_request[offset:])
         buffer.truncate()
@@ -87,28 +96,24 @@ def validate_request(request: Mapping[Any, Any]) -> None:
         raise ValidationError("Missing 'method' key")
     if "params" in request:
         if not isinstance(request["params"], list):
-            raise ValidationError("Missing 'method' key")
+            raise ValidationError(
+                f"The `params` value must be a list.  Got: {type(request['params'])}"
+            )
 
 
-def generate_response(
-    request: RPCRequest, result: Any, error: Optional[str]
-) -> RPCResponse:
+def generate_error_response(request: RPCRequest, error: str) -> RPCResponse:
     response = RPCResponse(
-        id=request.get("id", -1), jsonrpc=request.get("jsonrpc", "2.0"),
+        id=request.get("id", -1),
+        jsonrpc=request.get("jsonrpc", "2.0"),
+        error=str(error),
     )
+    return response
 
-    if result is None and error is None:
-        raise TypeError("Must supply either result or error for JSON-RPC response")
-    if result is not None and error is not None:
-        raise TypeError(
-            "Must not supply both a result and an error for JSON-RPC response"
-        )
-    elif result is not None:
-        response["result"] = result
-    elif error is not None:
-        response["error"] = str(error)
-    else:
-        raise Exception("Unreachable code path")
+
+def generate_success_response(request: RPCRequest, result: Any,) -> RPCResponse:
+    response = RPCResponse(
+        id=request.get("id", -1), jsonrpc=request.get("jsonrpc", "2.0"), result=result,
+    )
 
     return response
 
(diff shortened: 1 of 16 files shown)


ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+
+    while True:
+        data = await socket.recv(1024)
+        buffer.write(data.decode())
+
+        bad_prefix, raw_request = strip_non_json_prefix(buffer.getvalue())
+        if bad_prefix:
+            logger.info("Client started request with non json data: %r", bad_prefix)
+            await write_error(socket, f"Cannot parse json: {bad_prefix}")
+            continue
```
lithp, line 46: I'm not sure in which situations the client will give us non-JSON data, but it seems unlikely that we'll receive any more requests which make sense, this might want to kill the connection!

And if it doesn't kill the connection, I think it should clear the buffer, right now the buffer is left as-is, so we'll loop and throw the same failure every time we receive more data.

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+
+        try:
+            request, offset = decoder.raw_decode(raw_request)
+        except json.JSONDecodeError:
```
lithp, line 50: I'm not sure what a better alternative would be, but `JSONDecodeError` is thrown in more situations than an incomplete object. If we get truly malformed JSON this will also loop forever, disconnecting might be better UX for the other side of this connection.

lithp: Maybe checking that All '{'s have a matching '}'? It'd be nice if there was a way to do that without writing a JSON parser :thinking:

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+        raise ValidationError("Missing 'jsonrpc' key") from err
+    else:
+        if version != "2.0":
+            raise ValidationError(f"Invalid version: {version}")
+
+    if "method" not in request:
+        raise ValidationError("Missing 'method' key")
+    if "params" in request:
+        if not isinstance(request["params"], list):
+            raise ValidationError("Missing 'method' key")
```
lithp, line 90: `s/Missing 'method' key/'params' key must be a list/`

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def generate_response(
```
lithp, line 93: Instead of a single method which uses `TypeError` to manually implement polymorphism, this would better lean on the language if it was two separate methods, `generate_success_response` and `generate_error_response`

lithp: Oh, `RPCHandler` does this :sweat_smile: having the helpers up there means you wouldn't have to re-implement them below, `RPCHandler` would be a little cleaner.

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+    async def __call__(self, request: RPCRequest) -> RPCResponse:
+        try:
+            params = self.extract_params(request)
+            result = await self.do_call(params)
+        except RPCError as err:
+            return self.generate_error_response(request, str(err))
+        else:
+            return self.generate_success_response(request, result)
+
+    @abstractmethod
```
lithp, line 134: *nit*: If someone reads through this class this will already be pretty clear, it's just 6 lines up. But to be more explicit, it could be useful to include a docstring here explaining that you should throw `RPCError` if you want your error to be returned to the caller.

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+
+    async def wait_serving(self) -> None:
+        await self._serving.wait()
+
+    async def run(self) -> None:
+        self.manager.run_daemon_task(self.serve, self.ipc_path)
+        try:
+            await self.manager.wait_finished()
+        finally:
+            self.ipc_path.unlink()
```
lithp, line 179: I'm not sure, but maybe this should use `missing_ok = True`? If `self.serve` fails for some reason, maybe it couldn't bind to the given path, then we'll get an exception, and we'll also get a second exception here which might mask the first.

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+            self.logger.debug("Error handling request: %s", request, exc_info=True)
+            response = generate_response(request, None, f"Unexpected Error: {err}")
+        finally:
+            return json.dumps(response)
+
+    async def serve(self, ipc_path: pathlib.Path) -> None:
+        self.logger.info("Starting RPC server over IPC socket: %s", ipc_path)
+
+        with trio.socket.socket(trio.socket.AF_UNIX, trio.socket.SOCK_STREAM) as sock:
+            # TODO: unclear if the following stuff is necessary:
```
lithp, line 200: [The man page](https://linux.die.net/man/2/bind) is a little vague on this, but I'm pretty sure `SO_REUSEADDR` has no effect on unix domain sockets. [net/unix/af_unix.c](https://github.com/torvalds/linux/blob/master/net/unix/af_unix.c) never touches `sk_reuse` before deciding to send `EADDRINUSE`.

ddht/rpc.py
```diff
@@ -0,0 +1,253 @@
...
+                    result = await self.execute_rpc(cast(RPCRequest, request))
+                except Exception as e:
+                    self.logger.exception("Unrecognized exception while executing RPC")
+                    await write_error(socket, "unknown failure: " + str(e))
+                else:
+                    if not result.endswith(NEW_LINE):
+                        result += NEW_LINE
+
+                    try:
+                        await socket.send(result.encode())
```
lithp, line 251: I don't know what the correct thing to do is, but I notice this is a naked `encode()`, while [above](https://github.com/ethereum/ddht/pull/90/files#diff-b3928bcb6f44d42e17823787bdeabc5aR74) the code explicitly uses utf-8. Are these intentionally inconsistent?

ddht/rpc_handlers.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        return stats
+
+    @to_dict
+    def _bucket_stats(self) -> Iterator[Tuple[int, BucketInfo]]:
+        buckets_and_replacement_caches = zip(
+            self._routing_table.buckets, self._routing_table.replacement_caches,
+        )
+        for idx, (bucket, replacement_cache) in enumerate(
+            buckets_and_replacement_caches, 1
```
lithp, line 49: *nit*: I had to look up what this parameter is for, passing `start=1` is slightly more clear

ddht/v5_1/app.py
```diff
@@ -102,13 +93,19 @@ async def run(self) -> None:
...
         network = Network(client=client, bootnodes=bootnodes,)
 
-        logger.info("Protocol-Version: %s", self._boot_info.protocol_version.value)
-        logger.info("DDHT base dir: %s", self._boot_info.base_dir)
-        logger.info("Starting discovery service...")
-        logger.info("Listening on %s:%d", listen_on, port)
-        logger.info("Local Node ID: %s", encode_hex(enr_manager.enr.node_id))
-        logger.info(
+        if self._boot_info.is_rpc_enabled:
+            handlers = get_core_rpc_handlers(network.routing_table)
```
lithp, line 97: This is missing the `enr` parameter.

ddht/tools/web3.py
```diff
@@ -0,0 +1,85 @@
...
+from eth_utils import decode_hex
+from web3.method import Method
+from web3.module import ModuleV2
+from web3.types import RPCEndpoint
+
+from ddht.rpc_handlers import BucketInfo as BucketInfoDict
+from ddht.rpc_handlers import NodeInfoResponse, TableInfoResponse
+
+
+class NodeInfo(NamedTuple):
```
lithp, line 20: No strong preference, but in the interests of removing duplication it might be better to add these `from_rpc_response` methods to the original classes back in `rpc_handlers`. These classes are duplicates which must always be kept in sync with the originals!

lithp: Though, wouldn't it be cool if there was a json library which took a `NamedTuple` like this and populated it from the received json object, throwing errors for missing or mistyped fields. If these classes were any more complicated I'd look into using [marshmallow](https://marshmallow.readthedocs.io/en/stable/).

ddht/tools/web3.py
```diff
@@ -0,0 +1,85 @@
...
+            },
+        )
+
+
+class RPC:
+    nodeInfo = RPCEndpoint("discv5_nodeInfo")
+    routingTableInfo = RPCEndpoint("discv5_routingTableInfo")
+
+
+# TODO: why does mypy think ModuleV2 is of `Any` type?
```
lithp, line 77: This is annoying :thinking: I wonder if we have to put something into `mypy.ini` to tell it where to find types, or to tell it that it should be looking for types.

lithp (approved):
Left some comments but on the whole this looks good

ddht/rpc.py
pipermerriam: So we have to do it this way.  We could have read only the first part of a json string and the rest hasn't come across the socket yet.  Hence, optimistic parsing, and then continuing to try and read more from the socket and see if we can parse it again.

This could be improved to do something like enforce a maximum request payload size quite easily.... which would at least get us out of the infinite loop so I'll do that.

ddht/tools/web3.py
pipermerriam: They are implemented here so that the web3 specific stuff doesn't leak into the other RPC stuff.  Yes they have to be kept inline, but they are also all strongly typed and mypy tells us right away when they diverge (in most cases....).  Unless you feel strongly.  Also, `TypedDict` doesn't allow you to add methods.  It is pure typing sugar.

ddht/v5_1/app.py
pipermerriam: Looks like the `to_dict` decorator isn't preserving the function arguments... :cry:

PR #92, opened 2020-09-22
v5.1 Specification Update

pipermerriam:
## What was wrong?

https://github.com/ethereum/devp2p/pull/157

The v5.1 spec is now up to date.  Our implementations needs to be updated to match it.

## How was it fixed?

Updated it.

#### Cute Animal Picture

snuzzy-dot-com


pipermerriam:
This *partially* gets the v5.1 implementation up-to-date with the spec.  There is still at least one more update that needs to happen with the `id_nonce_signature`

pipermerriam:
Ok, this has successfully interop'd with the Nim and Go implementations.  That gives me adequate confidence to get this merged and iterate from there.

ddht/cli_parser.py
```diff
@@ -104,12 +104,21 @@ def __call__(
 network_parser.add_argument(
     "--listen-address", type=ipaddress.ip_address, help="IP address to listen on"
 )
-network_parser.add_argument(
+
+bootnodes_parser_group = network_parser.add_mutually_exclusive_group()
+bootnodes_parser_group.add_argument(
     "--bootnode",
     action=NormalizeAndAppendENR,
     help="IP address to listen on",
```
lithp, line 112: I think this helpstring is incorrect

ddht/v5_1/app.py
```diff
@@ -106,11 +106,12 @@ async def run(self) -> None:
         self.logger.info("Protocol-Version: %s", self._boot_info.protocol_version.value)
         self.logger.info("DDHT base dir: %s", self._boot_info.base_dir)
         self.logger.info("Starting discovery service...")
-        self.logger.info("Listening on %s:%d", listen_on, port)
```
lithp, line 109: why was the port dropped? that seems like it could be useful!

lithp (approved):
I didn't read through it carefully enough to find bugs, but read through it and tried to find anything which was confusing. Left some tiny comments but this looks good.

ddht/v5_1/app.py
pipermerriam: Not dropped.  It was being double reported.  `listen_on` is an `Endpoint` which contains both IP and port in a single object.

PR #94, opened 2020-09-24
Piper/implement json rpc ping endpoint

pipermerriam:
## What was wrong?

Need more JSON-RPC endpoints to allow interaction with the network via JSON-RPC

## How was it fixed?

Added a `discv5_ping` endpoint for sending a ping.

#### Cute Animal Picture

lionbear

### DIFF
--- a/ddht/tools/web3.py
+++ b/ddht/tools/web3.py
@@ -1,4 +1,7 @@
-from typing import Callable, Mapping, NamedTuple, Tuple
+import ipaddress
+from typing import Any, Callable, List, Mapping, NamedTuple, Tuple, Union
+
+from eth_utils import add_0x_prefix, encode_hex, remove_0x_prefix
 
 try:
     import web3  # noqa: F401
@@ -7,7 +10,7 @@ except ImportError:
 
 
 from eth_enr import ENR, ENRAPI
-from eth_typing import NodeID
+from eth_typing import HexStr, NodeID
 from eth_utils import decode_hex
 from web3.method import Method
 from web3.module import ModuleV2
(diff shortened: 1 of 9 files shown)


ddht/tools/web3.py
```diff
@@ -69,17 +74,77 @@ def from_rpc_response(cls, response: TableInfoResponse) -> "TableInfo":
...
 class RPC:
     nodeInfo = RPCEndpoint("discv5_nodeInfo")
     routingTableInfo = RPCEndpoint("discv5_routingTableInfo")
 
+    ping = RPCEndpoint("discv5_ping")
+
+
+def ping_munger(
+    module: Any, identifier: Union[ENRAPI, str, bytes, NodeID, HexStr]
+) -> List[str]:
```
lithp, line 100: low priority but it took me a bit to figure out why this was returning a list. A short description, or a link to https://github.com/ethereum/web3.py/blob/master/web3/method.py#L77 would have saved me  a couple minutes, and might save the time of a hypothetical future reader.

pipermerriam: cc @kclowes or @marcgarreau 

This is an area of the web3.py internals that *could* be documented at some point assuming we want to encourage people to extend the library like this.  I struggled a bit to figure out how to write a `Method(...)` properly and only got there by futzing and copy/pasta from the web3.py source code.

lithp (approved):
No complaints here.

ddht/tools/web3.py
kclowes: Yeah, that's a good idea! We can definitely document it. I think it's generic enough that we could allow people to extend with a minimal amount of effort. Added issue ethereum/web3.py#1755

PR #95, opened 2020-09-24
WIP: Alexandria

pipermerriam:
## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -90,6 +90,15 @@ class BroadcastLogAPI(ABC):
         ...
 
 
+class ContentValidatorAPI(ABC):
+    @abstractmethod
+    async def validate_content(self, content_key: ContentKey, content: bytes) -> None:
+        ...
+
+    async def validate_header(self, content_key: ContentKey, content: bytes) -> None:
+        ...
+
+
 class ContentProviderAPI(ServiceAPI):
     @abstractmethod
     async def ready(self) -> None:
@@ -416,6 +425,8 @@ class AlexandriaNetworkAPI(ServiceAPI, TalkProtocolAPI):
     advertisement_db: AdvertisementDatabaseAPI
     advertisement_provider: AdvertisementProviderAPI
     advertisement_manager: AdvertisementManagerAPI
+
+    content_validator: ContentValidatorAPI
 
     content_storage: ContentStorageAPI
     content_provider: ContentProviderAPI
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -167,6 +167,14 @@ class AdvertisementDatabaseAPI(ABC):
         ...
 
 
+class ContentManagerAPI(ServiceAPI):
+    content_storage: ContentStorageAPI
+
+    @abstractmethod
+    async def process_content(self, content_key: ContentKey, content: bytes) -> None:
+        ...
+
+
 class AdvertisementManagerAPI(ServiceAPI):
     new_advertisement: EventAPI[Advertisement]
 
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -7,6 +7,7 @@ from typing import (
     Iterable,
     Optional,
     Sequence,
+    Sized,
     Tuple,
     Type,
     Union,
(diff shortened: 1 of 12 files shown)


pipermerriam:
This has been partially taken apart into smaller PRs now.  See #110 #109 #108 #107 #106 

They each need to be individually cleaned up because they are all a little broken from the process of untangling them from each other.

pipermerriam:
The last commit in this can still actually be taken apart into two separate pieces which I *think* is going to make all of these PRs within the reviewable size (<300 SLOC)

pipermerriam (commented):
@lithp I'm curious to get your high level feedback on the latest commit here: a63138cd33bf4f2ef7932e0fe184cd18561b8a1f which converts the logic for responding to ping messages into a generic service that can be used in both alexandria and the base protocol.

It may not seem justified for ping/pong logic but the real impetus for this is the FIND_NODES logic.

https://github.com/ethereum/ddht/blob/78a7afd3c37a4ac88fdf6c086772bbc6ce93bcc5/ddht/v5_1/network.py#L403-L450

That whole function body is duplicated almost 1:1 between the base protocol and alexandria.  Same for `NetworkAPI.recursive_find_nodes`.  The question is: does deduplication of this logic justify the typing complexity required to support it?

pipermerriam:
@lithp I went ahead and hammered out d83b341 which makes the code that serves FINDNODES requests generic.  It might be a better example to demonstrate the proposed change.

I'm not fully sold on this BTW.  I don't like copy/pasting the `NetworkAPI.find_nodes` and `NetworkAPI.recursive_find_nodes` but I'm also a little uneasy making them the same.... Both Alexandria and the base protocol currently share business logic for maintaining the routing table.  There are no guarantees that the business logic stays the same.

lithp:
My intuition is that we will probably end up preferring the way it was back in https://github.com/ethereum/ddht/commit/a63138cd33bf4f2ef7932e0fe184cd18561b8a1f . The implementations are the same now, and there might be some helper methods both will benefit from, but I expect we'll appreciate the extra flexibility when we want to make tweaks to how Alexandria works.

pipermerriam:
@lithp I don't follow.  a63138c implements `PingClientAPI` which co-locates the logic for ping/pong sending and the service for responding to ping message.  

You comment both sounds like you support a63138c 

> My intuition is that we will probably end up preferring the way it was back in a63138c 

And that  maybe you think we should keep them separate.

> but I expect we'll appreciate the extra flexibility when we want to make tweaks to how Alexandria works.

Clarification please :grinning:

lithp:
Hah, sorry, I misunderstood. I saw d83b34102ffb3f8cb3 and thought this was the first commit to start going down the route of merging these implementations. Didn't notice that https://github.com/ethereum/ddht/pull/95/commits/a63138cd33bf4f2ef7932e0fe184cd18561b8a1f was the first commit to start merging them.

I should have said https://github.com/ethereum/ddht/pull/95/commits/3043cfd93fb3f3237e277c2cd14fe4b6897980f7, I think I prefer the way is was back in https://github.com/ethereum/ddht/pull/95/commits/3043cfd93fb3f3237e277c2cd14fe4b6897980f7.

pipermerriam:
> I should have said 3043cfd, I think I prefer the way it was back in 3043cfd.

:+1: I'm dropping 04d259d (ping absttraction) and 9e8c65 (find nodes abstraction) into their own branch for reference and rolling them back from this branch.

pipermerriam:
Alright, running ahead now with copy/pasta of the remaining functionality needed for the `AlexandriaNetworkAPI` to properly manage the overlay routing table as well as CLI command for running a node on the alexandria network.

pipermerriam:
@lithp I've got an MVP of SSZ-based partial data proofs in place in this branch. Initial experiments suggest that with an upper bound of about 1200 bytes per datagram and taking advantage of as many optimizations for compressing the proofs over the wire as possible that we can get at least 768 bytes of actual data into each packet with the rest of the packet being taken up by merkle proof overhead.  I think we can probably move this number up higher to maybe 900 or 1000 bytes of data per packet if we take advantage of more aggressive compression like leaving out sections of the tree that are already proven by other packets.

Compression that I'm taking advantage of now:

- send the *length* custom encoded (saves about 30 bytes)
- omit padding nodes from the RHS of the tree (saves ~300 bytes)
- deduplication of the paths into the tree (2-3 bytes per tree node)
- clever bit packing for encoding the path for each node (2-3 bytes per tree node)

Other options that are still on the table are snappy compression which will probably squeeze out a few more bytes.  As it stands I'm going to reach a point by the end of the week where I start to polish this and carve off chunks for review, after which this serves as the foundation for setting up the messages for doing content transmission over alexandria which is really exciting.

PR #96, opened 2020-09-24
Fully implement TALK apis

pipermerriam:
## What was wrong?

Need both low level (client) and high level (network) APIs for the TALKREQ/TALKRESP messages.

## How was it fixed?

Finished implementing them with basic tests.

#### Cute Animal Picture

TELEMMGLPICT000124983942-xlarge_trans_NvBQzQNjv4BqyuLFFzXshuGqnr8zPdDWXiTUh73-1IAIBaONvUINpkg

### DIFF
--- a/ddht/v5_1/client.py
+++ b/ddht/v5_1/client.py
@@ -258,12 +258,12 @@ class Client(Service, ClientAPI):
         node_id: NodeID,
         *,
         protocol: bytes,
-        request: bytes,
+        payload: bytes,
         request_id: Optional[int] = None,
     ) -> int:
         with self._get_request_id(node_id, request_id) as message_request_id:
             message = AnyOutboundMessage(
-                TalkRequestMessage(message_request_id, protocol, request),
+                TalkRequestMessage(message_request_id, protocol, payload),
                 endpoint,
                 node_id,
             )
@@ -272,10 +272,10 @@ class Client(Service, ClientAPI):
         return message_request_id
 
     async def send_talk_response(
-        self, endpoint: Endpoint, node_id: NodeID, *, response: bytes, request_id: int,
+        self, endpoint: Endpoint, node_id: NodeID, *, payload: bytes, request_id: int,
     ) -> None:
         message = AnyOutboundMessage(
-            TalkResponseMessage(request_id, response,), endpoint, node_id,
+            TalkResponseMessage(request_id, payload), endpoint, node_id,
         )
         await self.dispatcher.send_message(message)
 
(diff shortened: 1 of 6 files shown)


ddht/v5_1/client.py
```diff
@@ -400,9 +400,17 @@ async def find_nodes(
...
     ) -> InboundMessage[TalkResponseMessage]:
-        raise NotImplementedError
+        with self._get_request_id(node_id) as request_id:
+            request = AnyOutboundMessage(
+                TalkRequestMessage(request_id, protocol, payload), endpoint, node_id,
+            )
+            async with self.dispatcher.subscribe_request(
+                request, TalkResponseMessage
+            ) as subscription:
+                with trio.fail_after(REQUEST_RESPONSE_TIMEOUT):
```
lithp, line 412: trio and context managers are both such beautiful additions to the language

lithp: There's some duplication of work here. `subscribe_request` calls `_manage_request_response` which has it's own call to `trio.move_on_after(REQUEST_RESPONSE_TIMEOUT)`. This will probably be confusing to the next person who tries to increase one of these timeouts and finds that the timeout has not increased.

I'm not entirely sure where it belongs:
- part of me thinks the timeout belongs in `_manage_request_response`, and that way there's a uniform timeout all requests use and callers don't need to worry about timeouts
- the other part of me thinks it belongs up here, and that the timeout in `_manage_request_response` ought to be removed. This gives callers the flexibility to change the timeout and I have the feeling they'll sometimes want to. It also makes things more explicit for someone browsing the code and who wonders what `talk` does and how it handles requests which come in too late.

lithp (approved):
The `subscribe_request` and `_manage_request_response` code looks tricky enough that a test that the timeout works might also be in order, and I left a comment re duplicated timeouts, but this looks great to me!

ddht/v5_1/client.py
pipermerriam: I'll look into seeing if the timeout in `_manage_request_response`, which will result in the channel being closed, will cause the call on the next line to `await subscription.receive()` to throw something like `trio.BrokenResourceError` which can then either be allowed to bubble, or potentially we can raise an internal error of some kind.

pipermerriam: Alright.  I was able to remove all of these second-layer timeouts and rely solely on the internal timeout that arises from `_manage_request_response`.  Now these methods raise `trio.EndOfChannel` which I think we'll want to eventually replace with a custom internal exception.  I originally had it re-raising `trio.TooSlowError` but I reverted that since it's possible that the channel gets closed due to some other mechanism, so eventually we may want to re-raise an internal exception that is more generic like `RequestFailure`

Issue #99, opened 2020-10-06
Decompose `NetworkAPI` long running processes


pipermerriam:
_If this is a bug report, please fill in the following sections.
If this is a feature request, delete and describe what you would like with examples._

## What was wrong?

The network API has multiple long running processes for doing things like:

- responding to PING with PONG
- answering FINDNODES
- pinging the *oldest* routing table entry

By bundling these together, it makes testing their individual functionality difficult.

## How can it be fixed?

Decompose these into their own independent services.  Test them in isolation.  Run them as child services in the `NetworkAPI`

pipermerriam:
This was implemented in https://github.com/ethereum/ddht/pull/137 and we decided it wasn't worth it.

PR #100, opened 2020-10-06
Decompose ABC interfaces into more modular units

pipermerriam:
## What was wrong?

While working on #95 it became clear that implementing the overlay network is going to involve mirroring a lot of the functionality from the base implementation.  Specifically all of the logic surrounding management of the routing table (PING, PONG, FIND_NODES, NODES).

For example, #99 will split out some of this functionality into re-usable services.  It should be possible for these services to be used in both the main network as well as the overlay network to avoid some duplication.

## How was it fixed?

WIP

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/ddht/abc.py
+++ b/ddht/abc.py
@@ -4,11 +4,13 @@ from typing import (
     TYPE_CHECKING,
     Any,
     AsyncContextManager,
+    Collection,
     Deque,
     Generic,
     Iterator,
     List,
     Optional,
+    Sequence,
     Tuple,
     Type,
     TypedDict,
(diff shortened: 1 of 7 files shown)


pipermerriam:
At the `Client` level I've pulled out two interfaces, `PingPongAPI` and `FindNodesAPI`

Next, I'll implement #99 to verify that we can make these units of business logic generic, and then in #95 I'll re-work the `AlexandriaAPI` to build from these base APIs.  This should validate that this approach is viable.

PR #107, opened 2020-10-08
Make request_id retrievable from the InboundMessage

pipermerriam:
## What was wrong?

The `InboundMessage` API is useful in Alexandria as well, however some of the mechanics of doing sub-protocol messages require that we can still access the `request_id` from the base protocol request/response.

## How was it fixed?

Allow explicitly specifying the `request_id` when instantiating `InboundMessage`.  This allows the base protocol `request_id` to  be preserved for sub-protocol messages.

#### Cute Animal Picture

unlikely-animal-friendships-13

### DIFF
--- a/ddht/base_message.py
+++ b/ddht/base_message.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Any, Generic, TypeVar
+from typing import Any, Generic, Optional, TypeVar
 
 from eth_typing import NodeID
 from eth_utils import int_to_big_endian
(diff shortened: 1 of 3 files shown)


lithp (approved):
It's a small change so I'm for it but I don't quite understand what this is for. When will `explicit_request_id` be set, and what will it be used for?

pipermerriam:
Precursor to the Alexandria implementation.

The `request_id` is implemented at the base protocol level, meaning that the messages in Alexandria don't have their own request id.  We still use `InboundMessage` to wrap the alexandria messages when feeding them through the subscription API.  If I'm subscribing to a `PingMessage` in Alexandria (not the base protocol message but the one defined within the Alexandria protocol), then we need access to the `request_id` in order to issue a correct response.  

I'll  document this better with a comment.

PR #108, opened 2020-10-08
Respond to unhandled TalkRequest with empty response

pipermerriam:
## What was wrong?

The protocol specifies that an unhandled `TALKREQ` should be responded to with an empty `TALKRESP`

## How was it fixed?

Implemented an API for registering sub-protocols.  Now the `NetworkAPI` will detect `TALKREQ` messages for sub-protocols that haven't been registered and will provide the empty response.

#### Cute Animal Picture

dog-orangatang

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -2,7 +2,7 @@ import itertools
 import logging
 import operator
 import secrets
-from typing import Collection, Iterable, List, Optional, Set, Tuple
+from typing import Collection, Dict, Iterable, List, Optional, Set, Tuple
 
 from async_service import Service
 from eth_enr import ENRAPI, ENRDatabaseAPI, ENRManagerAPI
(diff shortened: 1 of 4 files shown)


pipermerriam:
TODO: this still needs a test.

ddht/v5_1/network.py
```diff
@@ -143,7 +171,7 @@ async def talk(
     ) -> bytes:
         if endpoint is None:
             endpoint = self._endpoint_for_node_id(node_id)
-        response = await self.client.talk(endpoint, node_id, protocol, payload)
+        response = await self.client.talk(endpoint, node_id, protocol, payload,)
```
lithp, line 174: This one is a bit strange. Why are these kinds of updates happening in your recent PRs, did you update your version of a linter?

tests/core/v5_1/test_network.py
```diff
@@ -232,3 +234,53 @@ async def _do_talk_response(network):
...
+                    node_id=alice.node_id,
+                    protocol=ProtocolA.protocol_id,
+                    payload=b"payload-a",
+                    endpoint=alice.endpoint,
+                )
+            assert response == b""
+
+            alice_network.add_talk_protocol(ProtocolA())
+
+            with pytest.raises(trio.EndOfChannel):
```
lithp, line 280: Why should this return an `EndOfChannel`?

lithp (approved):
Have a question about the test but the rest looks good

ddht/v5_1/network.py
pipermerriam: Likely happened because it was originally short and fit on the line.  Then I added something that made `black` wrap the line.  Then I visually saw that it was missing a dangly comma and maybe added it manually.  Then I removed whatever extra stuff made the line have to be wrapped and `black` unwrapped it, but it didn't drop the dangly comma....

tests/core/v5_1/test_network.py
pipermerriam: Yeah, it's probably time for this to get addressed in the base protocol.  If you recall, you recently identified that I had duplicate timeouts being enforced on request/response.  In removing the outer timeout, the inner-timeout now manifests as a `trio.EndOfChannel`.  This is because request/response is implemented using a channel that the response is fed through.  When the inner timeout is triggered, the channel gets closed, which causes the part listening for the response to raise `trio.EndOfChannel`.

I think that we probably want to catch this internally and re-raise it as a `trio.TooSlowError`.

PR #109, opened 2020-10-08
Implement SubscriptionManagerAPI

pipermerriam:
## What was wrong?

The `DispatcherAPI` implements the subscription logic.  This logic is useful when designing sub-protocols, but isn't re-usable in the current form.

## How was it fixed?

Extracted the functionality to a standalone and generic implementation that isn't bound to the core protocol message types and can be extended for use in TALK subprotocols.

#### Cute Animal Picture

20-unlikely-animal-friends

### DIFF
--- a/ddht/subscription_manager.py
+++ b/ddht/subscription_manager.py
@@ -1,0 +1,73 @@
+import collections
+import logging
+from typing import Any, AsyncIterator, DefaultDict, NamedTuple, Optional, Set, Type
+
+from async_generator import asynccontextmanager
+from eth_typing import NodeID
+import trio
+
+from ddht.abc import SubscriptionManagerAPI
+from ddht.base_message import AnyInboundMessage, InboundMessage, TMessage
+from ddht.endpoint import Endpoint
+
+
+class _Subcription(NamedTuple):
+    send_channel: trio.abc.SendChannel[AnyInboundMessage]
+    filter_by_endpoint: Optional[Endpoint]
+    filter_by_node_id: Optional[NodeID]
+
+
+class SubscriptionManager(SubscriptionManagerAPI[TMessage]):
+    logger = logging.getLogger("ddht.SubscriptionManager")
+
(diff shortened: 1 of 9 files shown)


pipermerriam:
TODO: this needs standalone tests.

ddht/subscription_manager.py
```diff
@@ -0,0 +1,71 @@
...
+            InboundMessage[TMessage]
+        ](256)
+        subscription = _Subcription(send_channel, endpoint, node_id)
+        self.logger.debug("Subscription setup for: %s", message_type)
+        try:
+            self._subscriptions[message_type].add(subscription)
+            async with receive_channel:
+                yield receive_channel
+        finally:
+            self._subscriptions[message_type].remove(subscription)
```
lithp, line 71: nitpick: I'm pretty sure that as written line 67 can never fail (except with something crazy like `MemoryError`), but if it were to fail (maybe in the future the `defaultdict` is turned into something else) the finally block would also fail. The stacktrace would be a little easier to interpret if line 67 was outside of the `try:finally:` block.

ddht/subscription_manager.py
```diff
@@ -0,0 +1,71 @@
...
+                if message.sender_endpoint != subscription.filter_by_endpoint:
+                    continue
+            if subscription.filter_by_node_id is not None:
+                if message.sender_node_id != subscription.filter_by_node_id:
+                    continue
+
+            try:
+                subscription.send_channel.send_nowait(message)  # type: ignore
+            except trio.WouldBlock:
+                self.logger.debug(
```
lithp, line 46: Dropping these messages is a reasonable decision, but this behavior might deserve a warning message in a docstring!

lithp (approved):

ddht/subscription_manager.py
pipermerriam: I'll move it out because you are correct......  except for `KeyboardInterrupt` which can happen literally anytime (fun fact I learned from a blog post by the `trio` lead dev.

PR #110, opened 2020-10-08
Implement RequestTrackerAPI

pipermerriam:
## What was wrong?

The logic for tracking which request ids are in use is valuable when writing sub-protocols.  It was not re-usable in the current form, embedded int he `DispatcherAPI`

## How was it fixed?

Extracted the logic into a standalone `RequestTrackerAPI`

#### Cute Animal Picture

unlikely-animal-friendships-17

### DIFF
--- a/ddht/request_tracker.py
+++ b/ddht/request_tracker.py
@@ -1,0 +1,60 @@
+from contextlib import contextmanager
+import secrets
+from typing import Iterator, Optional, Set, Tuple
+
+from eth_typing import NodeID
+
+from ddht.abc import RequestTrackerAPI
+
+MAX_REQUEST_ID_ATTEMPTS = 3
+
+
+def _get_random_request_id() -> bytes:
+    return secrets.token_bytes(4)
+
+
+class RequestTracker(RequestTrackerAPI):
+    def __init__(self) -> None:
+        self._reserved_request_ids: Set[Tuple[NodeID, bytes]] = set()
+
+    def get_free_request_id(self, node_id: NodeID) -> bytes:
+        for _ in range(MAX_REQUEST_ID_ATTEMPTS):
+            request_id = _get_random_request_id()
+
+            if (node_id, request_id) in self._reserved_request_ids:
+                continue
+            else:
+                return request_id
+        else:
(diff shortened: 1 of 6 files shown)


pipermerriam:
TODO: this needs standalone tests.

ddht/request_tracker.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+
+    @contextmanager
+    def reserve_request_id(
+        self, node_id: NodeID, request_id: Optional[bytes] = None
+    ) -> Iterator[bytes]:
+        if request_id is None:
+            request_id = self.get_free_request_id(node_id)
+        try:
+            self._reserved_request_ids.add((node_id, request_id))
```
lithp, line 44: This doesn't throw an error if the `(node_id, request_id)` is already reserved, which is quite possible if the user passes `request_id` in manually. If that happens the finally block will throw an error, but it will be a very confusing error, which might not even be raised by the same coro which caused the problem!

I think this method needs an assertion checking that the request_id isn't already reserved.

lithp (approved):
Added a comment but conditional on that comment and maybe a test I'm all for this

ddht/request_tracker.py
pipermerriam: Yeah, I think this mildly violates the principle-of-least-surprise.  The API however is intentional.  I'll document it.

The reason is that there are *some* times in which `request_id` needs to be lazily generated, and other times when it needs to be supplied externally (typically by acquiring a clean one via `RequestTracker.get_free_request_id`.  The reason for the later is that within the Alexandria app, we need to be able to issue requests in a manner that we know the `request_id` in-flight.  This is the only way I could figure out how to do subscriptions for sub-protocol messages.  The `TALKRESP` message doesn't contain a `protocol_id`, only the `request_id` of the `TALKREQ` message (which does contain a `protocol_id`.  So, we have to track the active set of request ids for the sub-protocol so that we can match `TALKRESP` messages against known request ids to know that they are part of the sub-protocol and thus that they should be fed into the subscription.

Assuming the `request_id` is acquired through `get_free_request_id` the collision chances are very small, and even in the event of a collision, the consequences are very small.

Issue #93, opened 2020-09-24
Implement `crawl` command.


pipermerriam:
## What was wrong?

When the network is running it is useful to get a full picture of the network.

## How can it be fixed?

Implement `ddht crawl` which bootstraps into the network and then uses iterative `FIND_NODES` commands to walk the entire network and dump the found ENR records.

PR #111, opened 2020-10-09
Add a `crawl` command which attempts to traverse the network and log found ENRs

lithp:
Implements #93.

Starting at the bootnodes, recursively asks for the contents of each node's routing table and then asks the found nodes for *their* routing tables. Logs every found ENR; also logs which peers are and aren't responsive.

#### Cute Animal Picture

crawling-sloth

### DIFF
--- a/ddht/cli_commands.py
+++ b/ddht/cli_commands.py
@@ -1,7 +1,7 @@
 import logging
 import os
 
-from async_service import ServiceAPI, background_trio_service, as_service, Service
+from async_service import Service, ServiceAPI, as_service, background_trio_service
 import trio
 
 from ddht.boot_info import BootInfo
@@ -9,7 +9,6 @@ from ddht.constants import ProtocolVersion
 from ddht.v5.app import Application as ApplicationV5
 from ddht.v5.crawl import Crawler
 from ddht.v5_1.app import Application as ApplicationV5_1
-
 
 logger = logging.getLogger("ddht")
 
@@ -26,6 +25,7 @@ async def do_main(boot_info: BootInfo) -> None:
     logger.info("Started main process (pid=%d)", os.getpid())
     async with background_trio_service(app) as manager:
         await manager.wait_finished()
+
 
 async def do_crawl(boot_info: BootInfo) -> None:
 
--- a/ddht/cli_commands.py
+++ b/ddht/cli_commands.py
@@ -9,11 +9,16 @@ from ddht.constants import ProtocolVersion
 from ddht.v5.app import Application as ApplicationV5
 from ddht.v5_1.app import Application as ApplicationV5_1
 
+from ddht.v5.app import get_local_private_key
+from eth_enr import ENRDB, default_identity_scheme_registry, ENRManager, ENR
+from ddht.base_message import AnyInboundMessage, AnyOutboundMessage
 from ddht.constants import DEFAULT_LISTEN
-from ddht.datagram import send_datagram, InboundDatagram, DatagramReceiver
-from ddht.v5.channel_services import InboundPacket, PacketDecoder
+from ddht.datagram import send_datagram, InboundDatagram, DatagramReceiver, OutboundDatagram, DatagramSender
+from ddht.v5.channel_services import InboundPacket, PacketDecoder, OutboundPacket, PacketEncoder
 from ddht.v5.constants import DEFAULT_BOOTNODES
-from ddht.v5.messages import PingMessage
+from ddht.v5.messages import PingMessage, v5_registry
+from ddht.v5.message_dispatcher import MessageDispatcher
+from ddht.v5.packer import Packer
 
 logger = logging.getLogger("ddht")
 
@@ -51,20 +56,28 @@ async def do_listen(boot_info: BootInfo) -> None:
     )
 
     listen_on = boot_info.listen_on or DEFAULT_LISTEN
+
+    enr_db = ENRDB(dict(), default_identity_scheme_registry)
+    local_private_key = get_local_private_key(boot_info)
+    enr_manager = ENRManager(private_key=local_private_key, enr_db=enr_db)
+    enr_manager.update((b"udp", boot_info.port))
+    enr_manager.update((b"ip", listen_on.packed))
+
     logger.info(f"About to listen. bind={listen_on}:{boot_info.port}")
+    logger.info(f"current enr. enr={enr_manager.enr}")
     await sock.bind((str(listen_on), boot_info.port))
 
     async def log_received_packets():
         async for packet in inbound_packet_channels[1]:
             logger.info(f"Received packet from {packet.sender_endpoint}: {packet}")
 
+    # TOOD: is there a way to do this which involves less ceremony?
+    #       I think the core benefit async_service provides is that it's trio & asyncio?
+
     class ListenService(Service):
         async def run(self) -> None:
             logger.info("starting listen service")
-            services = [
-                datagram_receiver, packet_decoder, # as_service(log_received_packets)
-            ]
-            for service in services:
+            for service in (datagram_receiver, packet_decoder):
                 self.manager.run_daemon_child_service(service)
             self.manager.run_daemon_task(log_received_packets)
             await self.manager.wait_finished()
@@ -90,14 +103,13 @@ async def do_crawl(boot_info: BootInfo) -> None:
     sock = trio.socket.socket(
         family=trio.socket.AF_INET, type=trio.socket.SOCK_DGRAM
     )
-    await sock.bind(("0.0.0.0", 30306))
 
     # Bind to a socket and try to bond with a single bootnode
 
     test_bootnode = DEFAULT_BOOTNODES[0]
+    test_bootnode = "enr:-IS4QHoI1sHKaPmcfDU3m0KXBeoWK8uidDvvnUOy01oigUs3a0HQjx2U701KvBG0Dg5OAsU3A5f6W2CvdZEIGFXcxhgDgmlkgnY0gmlwhAAAAACJc2VjcDI1NmsxoQLgn7ZgocyVfD0cUxGpfI_flyR-Oe8GNqw1V9MeWwxKC4N1ZHCCdl8"
     logger.info(f"Attempting to bond with: {test_bootnode}")
-
-    ping = PingMessage(request_id=0, enr_seq=0)
+    bootnode_enr = ENR.from_repr(test_bootnode, default_identity_scheme_registry)
 
     # datagram = packet.to_wire_bytes()
     # send_datagram(sock, datagram, endpoint)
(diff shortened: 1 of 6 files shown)


pipermerriam (commented):
forgive my preliminary review.  looks like it's going in a good direction.

pipermerriam:
Also, I'm curious about the chosen algorithm.

Ask every node we see to return the nodes they know of in their furthest away buckets.  Once we've asked every node we assume we've crawled the network...

My intuition is that this will give us 99% of the network but that there's some decent probability that we miss nodes.  Bucket[256] represents the larges set of nodes, with 50% of the network being allocated to that bucket for any given node  That means that each node in the network will return 16 random nodes from 50% of the network.  I'm not going to do the math, but my intuition is that we'll regularly miss some.

I'm fine sticking with the current algorithm because I think it's "good enough", but it might be worth documenting the algorithm and its expected behavior.

One way to potentially improve this would be to do something like this:

```python
for bucket in range(256, 0, -1):
    found_nodes = find_nodes(node_id, bucket)

    if not found_nodes:
        break

    ... # process the found nodes
```

The idea being that we also want to query the *close* buckets for each node since nodes are expected to converge towards a *complete* view of their local part of the network.  This means that we get the vast majority of the nodes via querying bucket 256, and then we fill in the few ones that may have been missed by scraping the *close* buckets.

pipermerriam:
FYI: https://github.com/ethereum/ddht/pull/100 is one concept for how we *could* make this generic to either the v5 or v5.1 network.

pipermerriam:
You mentioned this was missing chunks of the network.  I'm curious to hear more detail on this and any intuition on why.  Big chuncks?  Small chunks?

Did you experiment with any other concurrency levels?  I'm wondering what our throughput levels can be and where the bottlenecks are.

ddht/v5/client.py
```diff
@@ -0,0 +1,96 @@
...
+class Client(Service):
+    logger = logging.getLogger("ddht.Client")
+
+    def __init__(
+        self, local_private_key: keys.PrivateKey, enr_db: ENRDatabaseAPI, node_id, sock
+    ) -> None:
+
+        self.enr_db = enr_db
+
+        outbound_datagram_channels = trio.open_memory_channel[OutboundDatagram](0)
```
pipermerriam, line 40: If you play with the concurrency levels or identifying bottlenecks, I would suggest tinkering with these channel buffer sizes.

lithp:
> You mentioned this was missing chunks of the network. I'm curious to hear more detail on this and any intuition on why. Big chuncks? Small chunks?

I turned the existing logic into a separate service called Phase 1, and then wrote a Phase 2 which loops forever, it picks a random node id and sends FindNodes to the nodes closest to that target id. Phase 1 finishes within ~ a minute and finds ~900 peers. I left it running overnight and it had discovered ~ 3000 peers when I came back to it. It was still occasionally finding ENRs which I think is the result of new nodes joining the network.

lithp:
> Did you experiment with any other concurrency levels? I'm wondering what our throughput levels can be and where the bottlenecks are.

Not too much, I played around with the concurrency parameter and found that 32 was able to hit every node in < 1 minute and decided that was good enough. I bet the bottleneck will be somewhere in our cryptography but I haven't tried to profile it.

lithp:
Well, I had originally gone with the idea of scanning just the last bucket because I thought that would get us close enough to finding every ENR. That turns out to have been quite wrong!

I just wrote something which does what you suggested, iterates through all the buckets for each node. It took about 10 times longer to complete (10 minutes in total) but it found 4378 different peers, 640 of them responsive.

pipermerriam:
I think that's a great starting point and we can whittle the efficiency and time down from there.

ddht/v5/crawl.py
```diff
@@ -0,0 +1,280 @@
...
+        )
+
+        self.private_key = get_local_private_key(boot_info)
+
+        self.enr_manager = ENRManager(private_key=self.private_key, enr_db=self.enr_db)
+        self.enr_manager.update((b"udp", boot_info.port))
+
+        # Setup all the services
+
+        outbound_datagram_channels = trio.open_memory_channel[OutboundDatagram](0)
```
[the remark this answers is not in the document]
lithp: Broke out the duplicated code into a `Client` class. `Client` doesn't currently expose the most rational interface but it's a good start.

ddht/v5/crawl.py
```diff
@@ -0,0 +1,280 @@
...
+        self.seen_nodeids = set()
+        self.bad_enr_count = 0
+
+    async def visit_enr(self, remote_enr):
+        task = get_current_task()
+
+        logger.info(f"[{task}] sending FindNode(256). nodeid={encode_hex(remote_enr.node_id)}")
+        try:
+            with trio.fail_after(2):
+                find_node = FindNodeMessage(request_id=0, distance=256)
```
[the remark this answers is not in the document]
lithp: Fixed this now that it's talking to more nodes, it now picks bettter request ids

ddht/v5/crawl.py
```diff
@@ -0,0 +1,280 @@
...
+                    enr_count = resp.message.total
+                    received_enrs = resp.message.enrs
+                    logger.info(f"[{task}] Received response. enr_count={enr_count}")
+
+                    for enr in received_enrs:
+                        await self.schedule_enr_to_be_visited(enr)
+
+                # we only send one packet per peer, so do some cleanup now or else we'll
+                # leak memory.
+                await self.packer.managed_peer_packers[remote_enr.node_id].manager.stop()
```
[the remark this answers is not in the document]
lithp: both are great points!

ddht/v5/crawl.py
```diff
@@ -0,0 +1,280 @@
...
+
+        successes = len(self.seen_nodeids) - self.bad_enr_count
+        logger.info(f"Finished crawling. found_enrs={len(self.seen_nodeids)} bad_enrs={self.bad_enr_count} successful_connects={successes}")
+
+
+class ENRQueue:
+    """
+    The script should quit when there is nothing left to crawl.
+
+    However, it can't quit when the ENR queue is empty, because the ENR queue might
```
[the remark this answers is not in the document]
lithp: Thanks for the prod to find something better. I had originally decided against doing that because I thought there would be race-conditions. However, we're running in a single thread! There are now some ugly `CAUTION` blocks to prevent future us from adding race-conditions back in but I think you're right that this approach is overall much cleaner.

lithp:
Currently looking into why the test is broken but I think this is ready for another review. There's still some cleanup left to be done but it'd also be useful to know what you think is in most need of some love.

ddht/cli_commands.py
```diff
@@ -23,3 +24,15 @@ async def do_main(boot_info: BootInfo) -> None:
     logger.info("Started main process (pid=%d)", os.getpid())
     async with background_trio_service(app) as manager:
         await manager.wait_finished()
+
+
+async def do_crawl(boot_info: BootInfo) -> None:
+
+    if boot_info.protocol_version is not ProtocolVersion.v5:
+        raise Exception("Currently crawling is only supported on the v5 network.")
```
pipermerriam, line 32: I think you *could* raise `argparse.ArgumentError` here.

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+from ddht.constants import DEFAULT_LISTEN, IP_V4_ADDRESS_ENR_KEY
+from ddht.exceptions import UnexpectedMessage
+from ddht.v5.app import get_local_private_key
+from ddht.v5.client import Client
+from ddht.v5.messages import FindNodeMessage
+
+logger = logging.getLogger("ddht.crawler")
+
+
+class StopScanning(Exception):
```
pipermerriam, line 22: Typically for this type of *internal* exception that is used in a closed-loop manner (it is raised and re-caught and should never make it out to userland), I've been naming them as `_StopScanning` to signal the internal nature.

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+                        peer_id
+                    ),
+                    distance=bucket,
+                )
+
+                responses = await self.client.message_dispatcher.request_nodes(
+                    peer_id, find_node
+                )
+
+                return [enr for response in responses for enr in response.message.enrs]
```
pipermerriam, line 72: For consistency with our other code this should probably be converted to a tuple comprehension `tuple(thing for thing in things)`

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+                # any peers in the given bucket decreases exponentially. we save some time
+                # by moving on to the next peer once we notice that the buckets have
+                # started to think out.
+                if consecutive_empty_buckets >= 5:
+                    return
+
+                for enr in enrs:
+                    await self.schedule_enr_to_be_visited(enr)
+
+                await trio.sleep(1)  # don't overburden this peer
```
pipermerriam, line 124: These requests should be pretty easy for peers to respond to.  Do you think we really need this?  Did you happen to explore how this effects performance?

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+        except StopScanning:
+            pass
+        finally:
+            # we only send one packet per peer. do some cleanup now or we'll leak memory.
```
pipermerriam, line 128: This comment is no longer applicable.

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+
+            # CAUTION: Do not insert any code here. There must not be any checkpoints
+            #          between popping from the queue and entering active_tasks.
+
+            with self.active_tasks.enter():
+                await self.scan_enr_buckets(enr)
+
+    async def schedule_enr_to_be_visited(self, enr: ENRAPI) -> None:
+        if enr.node_id in self.seen_nodeids:
+            # In order to be nicer on the network only send a single packet to each
```
pipermerriam, line 162: Is this comment still accurate?

ddht/v5/crawl.py
```diff
@@ -0,0 +1,234 @@
...
+
+    async def run(self) -> None:
+        logger.info("crawling!")
+
+        boot_info = self._boot_info
+
+        if boot_info.is_upnp_enabled:
+            logger.info(
+                "uPNP will not be used; crawling does not require listening for "
+                "incoming connections."
```
pipermerriam, line 194: Does UDP somehow magically punch through?  Even though we don't listen for connections we do have to receive responses to our messages.  Mostly just curious how this works because it confuses me.

pipermerriam (approved):
:+1:

ddht/v5/crawl.py
lithp: It sure does! I'm going off memory so might be getting this wrong, but this works in the same way that you can open TCP connections to sites like google.com without needing to punch any holes for them. When you send an outbound packet the NAT remembers the port number you used and for a short period of time forwards packets to that port to you.

If I'm reading [this](https://tools.ietf.org/html/rfc5382) right the port stays open for at least 4 minutes.

ddht/cli_commands.py
lithp: Turned it into a `ValueError`. I tried turning it into an `ArgumentError` but didn't have easy access to the parser its constructor wants.

ddht/v5/crawl.py
lithp: I changed it to .1s and it's certainly faster, it now runs in ~5.5 minutes, I'm okay with lowering it to there

lithp:
applied all your feedback, just need to figure out why the `help` test is failing on CI

pipermerriam:
test failure appears to be related to pyrlp shenanigans (doesn't have to block this PR if it is the only failure).  Rebasing on master might clean it up if you haven't done that already.

lithp:
nice call, fixed by a rebase, pyrlp must have been it

Issue #97, opened 2020-09-25
Argument order for ClientAPI `send_XXX` methods


pipermerriam:
## What was wrong?

The `NetworkAPI` uses the following pattern for it's send methods.

```python
def method(node_id: NodeID, endpoint: Endpoint, ...)
```

The `ClientAPI` is reversed

```python
def method(endpoint: Endpoint, node_id: NodeID)
```

This is confusing.

## How can it be fixed?

Change all of the methods that take both `node_id` and `endpoint` to take them in that order, with `node_id` first.

pipermerriam:
This was cleaned up in #113

PR #113, opened 2020-10-09
Standardize NetworkAPI and ClientAPI

pipermerriam:
## What was wrong?

Two semi-related things

- When building Alexandria it became apparent that specifying the request_id at the `NetworkAPI` level was necessary to be able to determine which incoming `TalkResponse` should be considered part of the protocol, by matching their `request_id` up against the set of currently in-flight request ids.
- #97  (endpoint, node_id) -> (node_id, endpoint)

## How was it fixed?

- Now both `NetworkAPI` and `ClientAPI` use a consistent argument order of `(node_id, endpoint)`
- Now all of the methods on both `NetworkAPI` and `ClientAPI` allow specifying the `request_id` for outgoing messages.

#### Cute Animal Picture

unlikely-animal-friendships-3


lithp (approved):

lithp:
This animal picture is extra cute

PR #114, opened 2020-10-13
Rename get_enr -> lookup_enr

lithp:
## What was wrong?

I use grep to navigate our codebases and having two methods called `get_enr` (`ENRDB.get_enr`) makes this codebase a little harder for me to navigate.

It feels weird for submit this because the right answer is to use better tools, but I don't think better tools exist for Python.

#### Cute Animal Picture

A tiny bird for a tiny PR

tiny-bird

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -192,7 +192,9 @@ class Network(Service, NetworkAPI):
         enrs = await self.find_nodes(node_id, 0, endpoint=endpoint)
         if not enrs:
             raise Exception("Invalid response")
-        # This reduce accounts for
+
+        # Assuming we're given enrs for a single node, this reduce returns the enr for
+        # that node with the highest sequence number
         return _reduce_enrs(enrs)[0]
 
     async def recursive_find_nodes(self, target: NodeID) -> Tuple[ENRAPI, ...]:
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -110,7 +110,7 @@ class Network(Service, NetworkAPI):
             return False
 
         try:
-            enr = await self.get_enr(node_id, enr_seq=pong.enr_seq, endpoint=endpoint)
+            enr = await self.lookup_enr(node_id, enr_seq=pong.enr_seq, endpoint=endpoint)
         except trio.EndOfChannel:
             self.logger.debug(
                 "Bonding with %s timed out during ENR retrieval",
@@ -171,7 +171,7 @@ class Network(Service, NetworkAPI):
         )
         return response.message.payload
 
-    async def get_enr(
+    async def lookup_enr(
         self, node_id: NodeID, *, enr_seq: int = 0, endpoint: Optional[Endpoint] = None
     ) -> ENRAPI:
         try:
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -16,7 +16,7 @@ import trio
 from ddht._utils import every, humanize_node_id
 from ddht.constants import ROUTING_TABLE_BUCKET_SIZE
 from ddht.endpoint import Endpoint
-from ddht.exceptions import DuplicateProtocol
+from ddht.exceptions import DuplicateProtocol, EmptyFindNodesResponse
 from ddht.kademlia import (
     KademliaRoutingTable,
     compute_distance,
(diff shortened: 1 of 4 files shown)


pipermerriam (approved):
I was kind of surprised there was no test in place for this already.  There is *some* internal logic that might be worth testing such as empty responses or the weird case where they return two valid ENR records with different sequence numbers.  I don't feel strongly that these tests are strictly needed, but they would be an improvement if you are up for hammering them out real quick.

lithp:
Added some tests and rebased onto master, I'll merge this tomorrow once you've had one last chance to look at it

lithp:
Hah, actually, after looking at why the tests are failing after the rebase, I think there might genuinely be a bug here!

Here's `lookup_enr`:

https://github.com/ethereum/ddht/blob/51882f0447dc3b7c724b0d75ed3fd31291afd6b4/ddht/v5_1/network.py#L174-L187

It first tries to return the enr from the `enr_db`, and if the enr isn't there it tries to send `FindNodes` to the peer. However, `find_nodes` fails if the peer is not in the database!

I'm not entirely sure what should happen here, I think the `KeyError` should probably be raised, and a request will only be made if you're asking for a higher sequence number than what is already known?

lithp:
Changed `lookup_enr` and got all the tests passing, if you choose to accept.

ddht/v5_1/network.py
```diff
@@ -171,18 +173,14 @@ async def talk(
...
         return response.message.payload
 
-    async def get_enr(
+    async def lookup_enr(
         self, node_id: NodeID, *, enr_seq: int = 0, endpoint: Optional[Endpoint] = None
     ) -> ENRAPI:
-        try:
-            enr = self.enr_db.get_enr(node_id)
-        except KeyError:
+        enr = self.enr_db.get_enr(node_id)
```
pipermerriam, line 179: So I think we want to revert to the previous logic and instead do it as:

```python
            enr = self.enr_db.get_enr(node_id)

        except KeyError:
            if endpoint is None:
                raise  # if we don't already know the `endpoint` it isn't possible for us to send a request.
            enr = await self._fetch_enr(node_id, endpoint=endpoint)
            self.enr_db.set_enr(enr)
```

The idea here is that we *can* issue the request if we happen to already know the `node_id` & `endpoint` and in that case, we have a path to success even if we have no entry in our database.

ddht/v5_1/network.py
```diff
@@ -192,7 +190,9 @@ async def _fetch_enr(
         enrs = await self.find_nodes(node_id, 0, endpoint=endpoint)
         if not enrs:
             raise Exception("Invalid response")
-        # This reduce accounts for
+
+        # Assuming we're given enrs for a single node, this reduce returns the enr for
```
pipermerriam, line 194: :heart:

tests/core/v5_1/test_network.py
```diff
@@ -100,6 +101,93 @@ async def test_network_find_nodes_api(alice, bob):
...
+                alice.endpoint,
+                enrs=[],
+                request_id=find_nodes.message.request_id,
+            )
+
+    with trio.fail_after(2):
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(return_empty_response)
+
+            with pytest.raises(Exception):
```
pipermerriam, line 133: Can we zero this `Exception` into something more specific?

tests/core/v5_1/test_network.py
```diff
@@ -100,6 +101,93 @@ async def test_network_find_nodes_api(alice, bob):
...
+    enr_manager = ENRManager(private_key=bob.private_key, enr_db=bob.enr_db,)
+    enr_manager.update(
+        (b"udp", bob.endpoint.port), (b"ip", bob.endpoint.ip_address),
+    )
+
+    first_enr = enr_manager.enr
+
+    enr_manager.update((b"eth2", b"\x00\x00"))
+
+    second_enr = enr_manager.enr
```
pipermerriam, line 170: You might make an assertion that you have two different sequence numbers here just for the sake of sanity checking.

pipermerriam (approved):
Thanks for adding the tests.

ddht/v5_1/network.py
lithp: ahh, great point. fixed

PR #116, opened 2020-10-14
Alexandria: Message and Payload types

pipermerriam:
## What was wrong?

Alexandria needs its own message types.

## How was it fixed?

Implemented Ping/Pong message types using SSZ


lithp (approved):
looks great

PR #117, opened 2020-10-14
Implement alexandria client

pipermerriam:
Builds from #116 

## What was wrong?

For Alexandria we need a *low level* API for message sending.

## How was it fixed?

Implemented `AlexandriaClientAPI` which implements the low level message sending for `PingMessage` and `PongMessage`

#### Cute Animal Picture

unlikely_animal_friendships_02

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -1,0 +1,143 @@
+from typing import Any, Optional, Set, Type
+
+from eth_typing import NodeID
+
+from ddht.endpoint import Endpoint
+from ddht.exceptions import DecodingError
+from ddht.request_tracker import RequestTracker
+from ddht.v5_1.abc import NetworkAPI
+from ddht.v5_1.alexandria.abc import AlexandriaClientAPI
+from ddht.v5_1.alexandria.constants import ALEXANDRIA_PROTOCOL_ID
+from ddht.v5_1.alexandria.messages import (
+    AlexandriaMessage,
+    PingMessage,
+    PongMessage,
+    TAlexandriaMessage,
+    decode_message,
+)
+from ddht.v5_1.alexandria.payloads import PingPayload, PongPayload
+
+
+class AlexandriaClient(AlexandriaClientAPI):
+    protocol_id = ALEXANDRIA_PROTOCOL_ID
+
+    _active_request_ids: Set[bytes]
+
+    def __init__(self, network: NetworkAPI) -> None:
+        self.network = network
+        self.request_tracker = RequestTracker()
+
+        self._active_request_ids = set()
+
+    #
+    # Request/Response message sending primatives
+    #
+    async def _send_request(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        message: AlexandriaMessage[Any],
+        *,
+        request_id: Optional[bytes] = None,
+    ) -> bytes:
+        data_payload = message.to_wire_bytes()
+        request_id = await self.network.client.send_talk_request(
+            node_id,
+            endpoint,
+            protocol=ALEXANDRIA_PROTOCOL_ID,
+            payload=data_payload,
+            request_id=request_id,
+        )
+        return request_id
+
+    async def _send_response(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        message: AlexandriaMessage[Any],
+        *,
+        request_id: bytes,
+    ) -> None:
+        data_payload = message.to_wire_bytes()
+        await self.network.client.send_talk_response(
+            node_id, endpoint, payload=data_payload, request_id=request_id,
+        )
+
+    async def _request(
+        self,
+        node_id: NodeID,
+        endpoint: Optional[Endpoint],
+        request: AlexandriaMessage[Any],
+        response_class: Type[TAlexandriaMessage],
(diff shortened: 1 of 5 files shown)


ddht/v5_1/alexandria/client.py
```diff
@@ -0,0 +1,130 @@
...
+    async def _send_request(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        message: AlexandriaMessage[Any],
+        *,
+        request_id: Optional[bytes] = None,
+    ) -> bytes:
+        data_payload = message.to_wire_bytes()
+        request_id = await self.network.client.send_talk_request(
```
lithp, line 44: *nit*: This looks a little strange, pulling out a member of a member breaks the law of demeter and makes this code a little more fragile during refactorings. Something here isn't quite factored right though I'm not sure what.

Looking through the rest of this class, it seems like it almost exclusively uses the `Client`. It calls `self.network.talk` but that's just a passthrough to the client. Maybe this class could be passed the client and the enr_manager?

ddht/v5_1/alexandria/client.py
```diff
@@ -0,0 +1,130 @@
...
+        # the Alexandria sub-protocol messages, we need to be able to identify
+        # which responses are part of the Alexandria sub-protocol.  The
+        # `request_id` is the mechanism we use to accomplish this.  We manually
+        # supply the `request_id` for requests so that we can use a separate
+        # `RequestTrackerAPI`.  Then, we can subscribe to **all**
+        # `TalkResponseMessage` messages, and only feed them through the
+        # subscription when the `request_id` is known by the tracker.
+        request_id = self.network.client.request_tracker.get_free_request_id(node_id)
+        request_data = request.to_wire_bytes()
+        with self.request_tracker.reserve_request_id(node_id, request_id):
```
lithp, line 84: It seems really strange that on 82 we use the client's request tracker, but here we use our own request tracker. Doesn't this mean that other other modules which talk to that client might try to use this request_id? We're only reserving it locally so they won't know not to use this request id.

lithp: If I'm reading this correctly `self.network.client.dispatcher.subscribe_request` already does dispatching based on the `request_id`, so this class shouldn't need to be worried about maintaining its own `request_id` namespace. Maybe I'm misreading something?

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+    async with bob_network.dispatcher.subscribe(TalkRequestMessage) as subscription:
+        async with trio.open_nursery() as nursery:
+            nursery.start_soon(_respond)
+
+            with trio.fail_after(1):
+                pong_message = await alexandria_client.ping(bob.node_id, bob.endpoint)
+                assert isinstance(pong_message, PongMessage)
+                assert pong_message.payload.enr_seq == bob.enr.sequence_number
```
lithp, line 63: *nit* it's definitely not necessary but it might be useful to include a test that sending a `Pong` with an incorrect request id does not cause `ping` to return.

ddht/v5_1/alexandria/client.py
pipermerriam: First, :heart: learning new terminology.  I wasn't aware of https://en.wikipedia.org/wiki/Law_of_Demeter

Depending on how strongly you feel about this. I'm -0 on fixing this mostly because I'm not sure how to without a lot of futzing against an API that is currently functioning very well.  The `ClientAPI` gives low level un-opinionated APIs for accessing the *raw* protocol.  The `NetworkAPI` exposes the business logic and hides away the internals and only exposes high level stuff.  What I found while building all of this out was that there were times when you needed each of these and that this separation makes testing easier and provides good encapsulation of these two tiers of functionality.

So let me know if either you feel strongly about this..

ddht/v5_1/alexandria/client.py
pipermerriam: The comment above was supposed to explain why this is happening but it appears I didn't do a very adequate job.

We need a subscription API for alexandria messages in order to be able to cleanly implement things like "send a pong anytime someone sends us a ping".  In order to do this we need to be able to monitor all TALKREQUEST and TALKRESPONSE messages for this protocol.  The TALKREQUEST message is easy since it contains a `protocol_id`.  For the TALKRESPONSE however, the only way for us to know that an incoming message belongs to this protocol is to match it against an in-flight request (which we implicitely know is part of the alexandria protocol).

So, this use of a separate `RequestTrackerAPI` is the way we can accomplish this.  We still acquire the `request_id` from the core request tracker (on line 82), so we get the guarantee that we are indeed using a unique `request_id`.  Then, we track that `request_id` in this Alexandria-specific tracker, so that we can monitor incoming TALKRESPONSE messages, check if their `request_id` is tracked, and if so: attempt to decode it as an alexandria message and feed it into any active subscriptions.

pipermerriam: I also recognize that this is probably more confusing since the functionality that is actually using this isn't introduced until #118 (the next PR in line)

pipermerriam: Ok, I updated the comment, I'm curious to get any feedback on how I might make that clearer.

PR #118, opened 2020-10-14
Implement alexandria network api

pipermerriam:
Builds on #116 and #117

## What was wrong?

Need a high level API for interacting with the Alexandria network

## How was it fixed?

Implemented `AlexandriaNetworkAPI`

This handles both high level message sending as well as the ability to subscribe to sub-protocol messages.

#### Cute Animal Picture

Unusual-animal-friendships-11

### DIFF
--- a/ddht/tools/driver/node.py
+++ b/ddht/tools/driver/node.py
@@ -12,6 +12,9 @@ from eth_utils import humanize_hash
 from ddht.endpoint import Endpoint
 from ddht.tools.driver.abc import NodeAPI
 from ddht.v5_1.abc import ClientAPI, EventsAPI, NetworkAPI
+from ddht.v5_1.alexandria.abc import AlexandriaNetworkAPI
+from ddht.v5_1.alexandria.client import AlexandriaClient
+from ddht.v5_1.alexandria.network import AlexandriaNetwork
 from ddht.v5_1.client import Client
 from ddht.v5_1.events import Events
 from ddht.v5_1.network import Network
@@ -74,3 +77,19 @@ class Node(NodeAPI):
         async with background_trio_service(network):
             await client.wait_listening()
             yield network
+
+    @asynccontextmanager
+    async def alexandria(
+        self, network: Optional[NetworkAPI] = None,
+    ) -> AsyncIterator[AlexandriaNetworkAPI]:
+        if network is None:
+            async with self.network() as network:
+                alexandria = AlexandriaNetwork(AlexandriaClient(network))
+                network.add_talk_protocol(alexandria)
+                async with background_trio_service(alexandria):
+                    yield alexandria
+        else:
+            alexandria = AlexandriaNetwork(AlexandriaClient(network))
+            network.add_talk_protocol(alexandria)
+            async with background_trio_service(alexandria):
+                yield alexandria
--- a/ddht/v5_1/alexandria/network.py
+++ b/ddht/v5_1/alexandria/network.py
@@ -1,0 +1,131 @@
+import logging
+from typing import AsyncContextManager, Optional, Type
+
+from async_service import Service
+from eth_enr import ENRManagerAPI
+from eth_typing import NodeID
+import trio
+
+from ddht.base_message import InboundMessage
+from ddht.endpoint import Endpoint
+from ddht.exceptions import DecodingError
+from ddht.subscription_manager import SubscriptionManager
+from ddht.v5_1.abc import NetworkAPI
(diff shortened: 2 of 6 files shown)


ddht/tools/driver/node.py
```diff
@@ -74,3 +77,19 @@ async def network(
         async with background_trio_service(network):
             await client.wait_listening()
             yield network
+
+    @asynccontextmanager
```
lithp, line 81: Weak preference for breaking this out into a standalone function, an `alexandria_network` context manager might be useful even if you aren't currently holding on to a `NodeAPI`. No objections to also adding a method to NodeAPI:

```python
async def alexandria(self):
  async with self.network() as network:
    async with alexandria_network(self.network) as alexandria:
       yield alexandria
```

I'm not sure what to call the standalone function though.

tests/core/v5_1/alexandria/test_subscriptions.py
```diff
@@ -0,0 +1,69 @@
...
+
+            message = await subscription.receive()
+
+        assert isinstance(message.message, PongMessage)
+        assert message.message.payload.enr_seq == alice.enr.sequence_number
+
+
+@pytest.mark.trio
+async def test_alexandria_subscription_via_talk_request_protocol_mismatch(
+    alice_network, alice, bob, bob_alexandria, autojump_clock
```
lithp, line 38: I'm starting to really enjoy magic.

ddht/v5_1/alexandria/network.py
```diff
@@ -0,0 +1,131 @@
...
+class AlexandriaNetwork(Service, AlexandriaNetworkAPI):
+    logger = logging.getLogger("ddht.Alexandria")
+
+    protocol_id = ALEXANDRIA_PROTOCOL_ID
+
+    _alexandria_client: AlexandriaClientAPI
+
+    def __init__(self, client: AlexandriaClientAPI) -> None:
+        self._alexandria_client = client
+        self.protocol_id = self._alexandria_client.protocol_id
```
lithp, line 34: *nit* it feels a little weird to set the `protocol_id` both here and in the top-level of the class.

It seems like this class, at the moment, makes almost no assumptions about what the protocol looks like, so maybe it should accept a protocol_id from the client.

If in the future it'll know more about the Alexandria protocol it might be better for this line to be an assert, that the client is using the same protocol we're using.

ddht/v5_1/alexandria/network.py
```diff
@@ -0,0 +1,131 @@
...
+            async for request in subscription:
+                await self._alexandria_client.send_pong(
+                    request.sender_node_id,
+                    request.sender_endpoint,
+                    enr_seq=self.enr_manager.enr.sequence_number,
+                    request_id=request.request_id,
+                )
+
+    async def _feed_talk_requests(self) -> None:
+        async with self.network.client.dispatcher.subscribe(
```
lithp, line 86: I like your reasoning on the other PR, I think you said that `network.client` present a unified interface. This looks like you're digging into internals but really the class has namespaced its methods for you for your convenience. That sounds great to me!

This line gives me more pause though :sweat_smile: This expands out into `self._alexandria_client.network.client.dispatcher`, which really feels like it's digging into implementation details. When any of these members are renamed or moved around the commit will need to also make unrelated changes: updating this class.

It's not a large harm, and I can't think of an easy way to fix it so if you also can't think of one I'm okay with merging this. Going forward I bet this could be simplified with the introduction of a message bus. Instead of needing to go and find the class which will let us subscribe to our messages we could listen for our messages.

lithp (approved):

ddht/tools/driver/node.py
pipermerriam: This `NodeAPI` is testing infrastructure, allowing us to bundle together all the different pieces that go into a node without needing to actually run one (`enr_db`, `enr_manager`, `node_id`, `private_key`, ...).  I am reasonably confident that there isn't a standalone use case for this outside of testing.

tests/core/v5_1/alexandria/test_subscriptions.py
pipermerriam: Being able to test timeouts without waiting for things to timeout FTW

ddht/v5_1/alexandria/network.py
pipermerriam: I'm going to mirror the subscribe API onto the `ClientAPI` which turns this into `alexandria_client.network.client.subscribe` which is still kind of *deep* into the object hierarchy, but it does eliminate the `DispatcherAPI` from the chain and further consolodates the *external* APIs onto the `ClientAPI` which is much more external facing than the dispatcher.

PR #120, opened 2020-10-16
Allow overiding payload instantiation for alexandria messages

pipermerriam:
Builds from #118 

## What was wrong?

While implementing `FINDNODES/FOUNDNODES` in alexandria I found that `ssz.List` types de-serialize to an internal representation of the list when we really just want a `tuple` of the values.

## How was it fixed?

Adjusted how Alexandria message payloads are initialized to allow for customization of the payload initialization to allow this kind of data munging.

#### Cute Animal Picture

image-6-for-unlikely-animal-friends-gallery-442675720


lithp (approved):
modulo my comments on the same code in the other PR, this looks great

PR #121, opened 2020-10-16
Convert alexandria client to be a service api

pipermerriam:
Builds from #121 

## What was wrong?

I was *hopeful* that the `AlexandriaClientAPI` could avoid being a `ServiceAPI`, however, while implementing `FINDNODES/FOUNDNODES` the need for subscriptions to be managed at that level arose.  This is so that we can get a multi-response from for the FOUNDNODES response.

## How was it fixed?

Moved the `AlexandriaNetworkAPI.subscription_manager` down to the `AlexandriaClientAPI` as well as the related functionality.  Also ended up needing to do some copy/pasta from `ddht.v5_1.dispatcher.Dispatcher` for the `subscribe_request` implementation.

#### Cute Animal Picture

unlikely-animal-friends-7

### DIFF
--- a/ddht/tools/driver/alexandria.py
+++ b/ddht/tools/driver/alexandria.py
@@ -1,0 +1,58 @@
+from typing import AsyncContextManager, AsyncIterator, Optional
+
+from async_generator import asynccontextmanager
+from async_service import background_trio_service
+
+from ddht._utils import asyncnullcontext
+from ddht.tools.driver._utils import NamedLock
+from ddht.tools.driver.abc import AlexandriaNodeAPI, NodeAPI
+from ddht.v5_1.abc import NetworkAPI
+from ddht.v5_1.alexandria.abc import AlexandriaClientAPI, AlexandriaNetworkAPI
+from ddht.v5_1.alexandria.client import AlexandriaClient
+from ddht.v5_1.alexandria.network import AlexandriaNetwork
+
+
+class AlexandriaNode(AlexandriaNodeAPI):
+    _lock: NamedLock
+
+    def __init__(self, node: NodeAPI) -> None:
+        self.node = node
+        self._lock = NamedLock()
+
+    @asynccontextmanager
+    async def client(
+        self, network: Optional[NetworkAPI] = None,
+    ) -> AsyncIterator[AlexandriaClientAPI]:
+        network_context: AsyncContextManager[NetworkAPI]
+
+        if network is None:
+            network_context = self.node.network()
+        else:
+            # unclear why the typing isn't work for `asyncnullcontext`
+            network_context = asyncnullcontext(network)  # type: ignore
+
+        async with self._lock.acquire("AlexandriaNode.client(...)"):
+            async with network_context as network:
+                alexandria_client = AlexandriaClient(network)
+                network.add_talk_protocol(alexandria_client)
+                async with background_trio_service(alexandria_client):
+                    yield alexandria_client
+
+    @asynccontextmanager
+    async def network(
+        self, network: Optional[NetworkAPI] = None,
+    ) -> AsyncIterator[AlexandriaNetworkAPI]:
+        network_context: AsyncContextManager[NetworkAPI]
+
+        if network is None:
+            network_context = self.node.network()
+        else:
+            # unclear why the typing isn't work for `asyncnullcontext`
+            network_context = asyncnullcontext(network)  # type: ignore
+
+        async with self._lock.acquire("AlexandriaNode.network(...)"):
+            async with network_context as network:
+                alexandria_network = AlexandriaNetwork(network)
+                network.add_talk_protocol(alexandria_network)
+                async with background_trio_service(alexandria_network):
+                    yield alexandria_network
--- a/ddht/tools/driver/node.py
+++ b/ddht/tools/driver/node.py
@@ -10,17 +10,18 @@ from eth_typing import NodeID
 from eth_utils import humanize_hash
 
 from ddht.endpoint import Endpoint
+from ddht.tools.driver._utils import NamedLock
 from ddht.tools.driver.abc import NodeAPI
+from ddht.tools.driver.alexandria import AlexandriaNode
 from ddht.v5_1.abc import ClientAPI, EventsAPI, NetworkAPI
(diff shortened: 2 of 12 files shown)


ddht/tools/driver/alexandria.py
```diff
@@ -0,0 +1,74 @@
...
+    _listen_lock: trio.Lock
+    _lock_owner: Optional[str]
+
+    def __init__(self, node: NodeAPI) -> None:
+        self.node = node
+        self._listen_lock = trio.Lock()
+        self._lock_owner = None
+
+    @asynccontextmanager
+    async def _acquire_listen_lock(self, owner: str) -> AsyncIterator[None]:
```
lithp, line 25: A comment would be useful here, I'm not sure what this lock is meant to protect. I think `self.node.network()`?

ddht/tools/driver/node.py
```diff
@@ -48,29 +57,45 @@ def endpoint(self) -> Endpoint:
     def node_id(self) -> NodeID:
         return self.enr.node_id
 
+    @asynccontextmanager
```
lithp, line 60: Now that this is used twice, it would save some lines for this to be in it's own class, `ListenLock`, which both of these callers can have an instance of.

ddht/v5_1/alexandria/client.py
```diff
@@ -110,6 +131,93 @@ async def _request(
...
+            "Sending request: %s with request id %s", request, request_id.hex(),
+        )
+
+        send_channel, receive_channel = trio.open_memory_channel[
+            InboundMessage[TAlexandriaMessage]
+        ](256)
+
+        with self.request_tracker.reserve_request_id(node_id, request_id):
+            async with trio.open_nursery() as nursery:
+                nursery.start_soon(
```
lithp, line 168: you should be able to do: `nursery.start_soon(self._manage_request_response, node_id, endpoint, ...)`

ddht/v5_1/alexandria/client.py
```diff
@@ -110,6 +131,93 @@ async def _request(
...
+    @asynccontextmanager
+    async def subscribe_request(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        request: AlexandriaMessage[Any],
+        response_message_type: Type[TAlexandriaMessage],
+        request_id: Optional[bytes],
+    ) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[TAlexandriaMessage]]]:
+        if request_id is None:
```
lithp, line 153: I think there might be a race condition if we ever add an `await` between the call to `get_free_request_id` and the call to `reserve_request_id`. reordering some of these blocks such that one happens right after the other means there's a slightly lower chance someone will add that `await` in the future.

ddht/v5_1/alexandria/client.py
```diff
@@ -110,6 +131,93 @@ async def _request(
...
+                        if response.request_id != request_id:
+                            continue
+                        else:
+                            await send_channel.send(response)
+        if scope.cancelled_caught:
+            self.logger.warning(
+                "Abandoned request response monitor: request=%s message_type=%s",
+                request,
+                response_message_type,
+            )
```
lithp, line 219: If I'm reading this right, won't this warning always be emitted? There's no `break` in the inner loop, it will keep trying to shuttle messages from one channel to another until it times out.

ddht/v5_1/alexandria/client.py
```diff
@@ -110,6 +131,93 @@ async def _request(
...
+
+    async def _manage_request_response(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        request: AlexandriaMessage[Any],
+        response_message_type: Type[AlexandriaMessage[Any]],
+        send_channel: trio.abc.SendChannel[InboundMessage[AlexandriaMessage[Any]]],
+        request_id: bytes,
+    ) -> None:
```
lithp, line 193: No recommendation now, but I bet this method gets simpler if `SubscriptionManager` supported allowed filtering by request_id, or timeouts, or if it accepted the channel it should be sending messages to. I bet that pulling messages from one channel and doing a little work and forwarding them to another will be common enough that spending some time to work out an interface which makes this method shorter is worth it.

lithp (approved):

ddht/v5_1/alexandria/client.py
pipermerriam: I'll add a comment to document this subtlty.

ddht/v5_1/alexandria/client.py
pipermerriam: Yeah, I think you're correct (and I'm seeing this in the base protocol as well)

ddht/v5_1/alexandria/client.py
pipermerriam: I'd be happy to clean any of this up.  It's all working pretty well right now but also, it's been developed in a vacuum with me as the only consumer of these APIs (and also having written them), so new :eyes: are appreciated.

ddht/v5_1/alexandria/client.py
pipermerriam: This was a workaround (which I'll document) for an inadequacy in how `trio` types the `start_soon` function.

```
ddht/v5_1/alexandria/client.py:178: error: Argument 1 to "start_soon" of "Nursery" has incompatible type "Callable[[NodeID, Endpoint, AlexandriaMessage[Any], Type[AlexandriaMessage[Any]], SendChannel[InboundMessage[AlexandriaMessage[Any]]], bytes], Coroutine[Any, Any, None]]"; expected "Union[Callable[[], Awaitable[Any]], Callable[[Any], Awaitable[Any]], Callable[[Any, Any], Awaitable[Any]], Callable[[Any, Any, Any], Awaitable[Any]], Callable[[Any, Any, Any, Any], Awaitable[Any]], Callable[[VarArg(Any)], Awaitable[Any]]]"
```

Anything beyond 4 args isn't covered in the type hints.

ddht/v5_1/alexandria/client.py
pipermerriam: Ok, no, not actually true.  So this function is run within a nursery.  When the outer nursery is cancelled, it doesn't trigger this because trio is smart and the cancellations are scoped such that they don't leak between cancel scopes.

PR #122, opened 2020-10-16
Implement alexandria find and found nodes client apis

pipermerriam:
## What was wrong?

The Alexandria network will need to maintain it's own overlay routing table containing only nodes that participate in the Alexandria protocol.  To do this we need the FINDNODES/FOUNDNODES message pairs and corresponding functionality.

## How was it fixed?

Implemented the base units for this functionality in `AlexandriaClientAPI`

#### Cute Animal Picture

unlikely-animal-friends

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -1,8 +1,8 @@
 from abc import abstractmethod
-from typing import Any, AsyncContextManager, Optional, Type
+from typing import Any, AsyncContextManager, Collection, Optional, Sequence, Tuple, Type
 
 from async_service import ServiceAPI
-from eth_enr import ENRDatabaseAPI, ENRManagerAPI
+from eth_enr import ENRAPI, ENRDatabaseAPI, ENRManagerAPI
 from eth_typing import NodeID
 import trio
 
@@ -12,6 +12,7 @@ from ddht.endpoint import Endpoint
 from ddht.v5_1.abc import NetworkAPI, TalkProtocolAPI
 from ddht.v5_1.alexandria.messages import (
     AlexandriaMessage,
+    FoundNodesMessage,
     PongMessage,
     TAlexandriaMessage,
 )
@@ -57,11 +58,44 @@ class AlexandriaClientAPI(ServiceAPI, TalkProtocolAPI):
     ) -> None:
         ...
 
+    @abstractmethod
+    async def send_find_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        *,
+        distances: Collection[int],
+        request_id: Optional[bytes] = None,
+    ) -> bytes:
+        ...
+
+    @abstractmethod
+    async def send_found_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        *,
+        enrs: Sequence[ENRAPI],
+        request_id: bytes,
+    ) -> int:
+        ...
+
     #
     # High Level Request/Response
     #
     @abstractmethod
-    async def ping(self, node_id: NodeID, endpoint: Endpoint,) -> PongMessage:
+    async def ping(self, node_id: NodeID, endpoint: Endpoint) -> PongMessage:
+        ...
+
+    @abstractmethod
+    async def find_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        distances: Collection[int],
+        *,
+        request_id: Optional[bytes] = None,
+    ) -> Tuple[InboundMessage[FoundNodesMessage], ...]:
         ...
 
 
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -1,15 +1,30 @@
 import functools
 import logging
-from typing import Any, AsyncContextManager, AsyncIterator, Optional, Set, Type
+from typing import (
+    Any,
+    AsyncContextManager,
+    AsyncIterator,
+    Collection,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+)
 
 from async_generator import asynccontextmanager
 from async_service import Service
+from eth_enr import ENRAPI
 from eth_typing import NodeID
+from eth_utils import ValidationError
 import trio
 
 from ddht.base_message import InboundMessage
 from ddht.endpoint import Endpoint
+from ddht.enr import partition_enrs
 from ddht.exceptions import DecodingError
+from ddht.kademlia import compute_log_distance
 from ddht.request_tracker import RequestTracker
 from ddht.subscription_manager import SubscriptionManager
 from ddht.v5_1.abc import NetworkAPI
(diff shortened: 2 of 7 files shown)


ddht/v5_1/client.py
```diff
@@ -389,6 +389,12 @@ async def find_nodes(
                     # `SessionTerminated` exception.
                     raise Exception("Invalid `total` counter in response")
 
+                # TODO: This validation likely belongs in the
+                # `NetworkAPI.find_nodes` method since the client API is
+                # supposed to be unopinionated about everything but the
+                # strictest validation for well formed messages.  This type of
+                # business-logic validation belongs higher in the stack.
```
pipermerriam, line 396: This is addressed in https://github.com/ethereum/ddht/pull/125

ddht/v5_1/alexandria/messages.py
```diff
@@ -79,6 +89,34 @@ class PongMessage(AlexandriaMessage[PongPayload]):
     payload: PongPayload
 
 
+@register
+class FindNodesMessage(AlexandriaMessage[FindNodesPayload]):
```
lithp, line 93: I know this isn't implementing v51 so the names don't have to match, but it seems extra confusing that `FindNodeMessage` is the v51 message and `FindNodesMessage` is the alexandria message, I can see myself mixing those two up a lot. I think calling this `FindNodeMessage` might be easier on future us.

ddht/v5_1/alexandria/abc.py
```diff
@@ -57,11 +58,44 @@ async def send_pong(
...
     #
     # High Level Request/Response
     #
     @abstractmethod
-    async def ping(self, node_id: NodeID, endpoint: Endpoint,) -> PongMessage:
+    async def ping(self, node_id: NodeID, endpoint: Endpoint) -> PongMessage:
+        ...
+
+    @abstractmethod
+    async def find_nodes(
```
lithp, line 91: Is there a chance you know why discv5 decided to change how `FindNodes` works? In v4 you send the target node id and the remote sends you the closest nodes, even if they're in different buckets. In v5 you ask for a specific bucket, and you have to write extra code to ask for additional buckets if it detects that few nodes were returned. v4 also allows you to target specific buckets by using specially crafted target node ids, I think the only advantage to v5 is that it allows you to target multiple non-contiguous buckets at once?

Depending on how you feel about this I would kind of prefer if Alexandria used the v4 approach. It seems like it's easier to program against and isn't any less powerful.

ddht/v5_1/alexandria/client.py
```diff
@@ -253,6 +275,38 @@ async def send_pong(
...
 
+    async def send_find_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        *,
+        distances: Collection[int],
+        request_id: Optional[bytes] = None,
+    ) -> bytes:
+        message = FindNodesMessage(FindNodesPayload(tuple(distances)))
```
lithp, line 286: I know I already approved this approach back in #116 but seeing this line is starting to give me some pause. Why are the payloads separate classes than the messages? Lines like this and methods like `from_payload_args` seem like they're workarounds to the decision to separate payloads from messages.

ddht/v5_1/alexandria/client.py
```diff
@@ -261,6 +315,55 @@ async def ping(self, node_id: NodeID, endpoint: Endpoint) -> PongMessage:
...
+        ) as subscription:
+            head_response = await subscription.receive()
+            total = head_response.message.payload.total
+            responses: Tuple[InboundMessage[FoundNodesMessage], ...]
+            if total == 1:
+                responses = (head_response,)
+            elif total > 1:
+                tail_responses: List[InboundMessage[FoundNodesMessage]] = []
+                for _ in range(total - 1):
+                    tail_responses.append(await subscription.receive())
```
lithp, line 341: I know this is just copying the existing implementation of `find_nodes`, so maybe the place to fix this is in a different PR, but this seems like this is a little too pessimistic. If the remote node sends us three `FoundNodes` messages but the last one is dropped we'll time out and return nothing to the caller, even though we received 2 messages with useful enrs.

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -82,3 +91,108 @@ async def _respond_wrong_request_id():
...
+                    remaining_messages.append(await subscription.receive())
+
+        all_decoded_messages = (decoded_first_message,) + tuple(
+            decode_message(message.message.payload) for message in remaining_messages
+        )
+        all_received_enrs = tuple(
+            itertools.chain(*(message.payload.enrs for message in all_decoded_messages))
+        )
+        expected_enrs_by_node_id = {enr.node_id: enr for enr in enrs}
+        assert len(expected_enrs_by_node_id) == len(enrs)
```
lithp, line 144: Since they're both sets, adding this assertion might actually make this test *harder* to debug. The test will fail here but won't give any indication of what has gone wrong. Comparing the sets directly, or letting the assert on 147 do all the work, will enable pytest to return better error messages

lithp (approved):
left some comments but nothing crucial, this looks good

ddht/v5_1/alexandria/messages.py
pipermerriam: I'd rather rename the v5 one because 1) we're likely to fully remove the v5 implementation at some point and 2) at least in v5.1, the plural feels more correct since in it's typical usage it finds multiple nodes.

ddht/v5_1/alexandria/abc.py
pipermerriam: I think that by sending the target you *leak* information about what you're looking for.  It appears the v5 approach makes some attacks harder.  cc @fjl who may be able to provide more information.

ddht/v5_1/alexandria/client.py
pipermerriam: Keeping the message and payload separate was intended to separate metadata from payload.

- metadata
    - message type flag
    - SSZ sedes to know how to encode/decode the payload
- payload
    - the actual data for the message

Admittedly, the metadata can be class-level properties since they don't change from message to message and the payloads are the dynamic parts that only exist at the instance level...

This design choice was more of a whim which was inspired by thinking certain things would be cleaner with the two separate.  I'm not sure it's better..  I'm ok unwinding it at some point if there's no clear benefit.  I'll open an issue to track this.

ddht/v5_1/alexandria/abc.py
fjl: In discv4, it was actually not possible to target any specific bucket because the parameter was the public key preimage. The actual target is hashed on the server side in v4. We decided to change this in v5 and eventually landed on the current approach. The discussion for this happened here: https://github.com/ethereum/devp2p/issues/79 https://github.com/ethereum/devp2p/issues/141

fjl: The rationale for using log distances in FINDNODE is that it's much harder for an attacker to short-circuit your lookup. The disadvantage is that routing is slightly slower.

ddht/v5_1/alexandria/client.py
pipermerriam: Opening an issue to track this as well.  Probably something we want to measure in practice and see how often we encounter this behavior.

ddht/v5_1/alexandria/messages.py
lithp: Okay, I'm also happy with the plural `FindNodesMessage`! I just expect it'll be easier on us if whichever name this uses it has the same name in every protocol.

ddht/v5_1/alexandria/abc.py
lithp: Thanks for the quick response, and for the links! This rationale makes sense to me, that's a nasty attack and avoiding it is pretty compelling.

PR #124, opened 2020-10-19
Alexandria: Network api for find nodes and add tests

pipermerriam:
builds on #122 

## What was wrong?

#122 implements the client APIs for doing FINDNODES/FOUNDNODES in Alexandria.  We still however need the high level APIs.

## How was it fixed?

Implement high level `AlexandriaNetworkAPI.find_nodes` and `AlexandriaNetworkAPI.recursive_find_nodes`.

Currently these are heavy on the code-duplication front.  After this has been merged, my plan is to go back and see about "drying" out these two independent implementations into something standardized.

This also implements tests for the `AlexandriaNetworkAPI` which were previously missing.

#### Cute Animal Picture

unusual-animal-friendship-cat-iguana__700

### DIFF
--- a/ddht/v5_1/alexandria/network.py
+++ b/ddht/v5_1/alexandria/network.py
@@ -1,15 +1,27 @@
+import itertools
 import logging
-from typing import Optional
+from typing import Collection, List, Optional, Set, Tuple
 
 from async_service import Service
-from eth_enr import ENRDatabaseAPI, ENRManagerAPI
+from eth_enr import ENRAPI, ENRDatabaseAPI, ENRManagerAPI
+from eth_enr.exceptions import OldSequenceNumber
 from eth_typing import NodeID
+from eth_utils.toolz import take
+import trio
 
+from ddht._utils import humanize_node_id, reduce_enrs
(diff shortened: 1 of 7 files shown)


pipermerriam:
@lithp this has now been rebased

ddht/v5_1/alexandria/network.py
```diff
@@ -64,6 +199,55 @@ async def _pong_when_pinged(self) -> None:
...
+        async with self.client.subscribe(FindNodesMessage) as subscription:
+            async for request in subscription:
+                response_enrs: List[ENRAPI] = []
+                distances = set(request.message.payload.distances)
+                if len(distances) != len(request.message.payload.distances):
+                    self.logger.debug(
+                        "Ignoring invalid FindNodesMessage from %s@%s: duplicate distances",
+                        humanize_node_id(request.sender_node_id),
+                        request.sender_endpoint,
+                    )
```
lithp, line 212: Makes sense but I wish we had a way of telling the other node what they did wrong. This seems like a weak violation of Postel's law, does the spec say we should ignore these messages?

lithp (approved):
looks good to me

PR #125, opened 2020-10-20
Move FINDNODES response validation into NetworkAPI

pipermerriam:
## What was wrong?

The `ClientAPI` is intended to be unopinionated.  This means allowing things that would typically go against the business logic but are within the rules of the networking protocol.

The current implementation of `ClientAPI.find_nodes` enforces validation against the distances of the ENR records returned.

## How was it fixed?

Moved the validation into `NetworkAPI.find_nodes`

#### Cute Animal Picture

unlikely_animal_friends-1920x1200


lithp (approved):
sounds reasonable to me, I like that distinction between client and network

pipermerriam:
> I like that distinction between client and network.

PR #128, opened 2020-10-21
Fix log level setup when using 0 as level

pipermerriam:
## What was wrong?

The logic used to determine log levels was not behaving correctly when the log level was set to `0`.

## How was it fixed?

Updated the logic to do an explicit `is None`.

#### Cute Animal Picture

ss-111014-unlikely-friends-elephant-sheep ss_full-1__880


pipermerriam:
Merging without review.  Happy to cleanup anything post merge if you have anything you'd like addressed.  Predicting this will be uncontroversial.

lithp (commented):
no complaints :+1:

PR #129, opened 2020-10-22
NetworkAPI.talk raises ProtocolNotSupported

pipermerriam:
## What was wrong?

The TALKREQ and TALKRESP protocol rules specify that an empty response is used to signal that the client does not support the protocol.  We should not treat these as valid responses, however `NetworkAPI.talk` would return them as the empty bytestring.

## How was it fixed?

Changed `NetworkAPI.talk` to raise a new `ddht.v5_1.exceptions.ProtocolNotSupported` exception in this case.

#### Cute Animal Picture

unlikely-animal-friendships-alliances-6

### DIFF
--- a/ddht/v5_1/exceptions.py
+++ b/ddht/v5_1/exceptions.py
@@ -1,2 +1,9 @@
-class SessionNotFound(Exception):
+from ddht.exceptions import BaseDDHTError
+
+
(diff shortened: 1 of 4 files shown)


lithp (approved):
looks great :+1:

PR #130, opened 2020-10-22
Use full hex NodeID in debug log messages

pipermerriam:
## What was wrong?

We were using the *humanized* NodeID in our DEBUG level log messages.  This isn't necessary since these messages are intended to be low level and should contain full information

## How was it fixed?

Use the full hex representation of the NodeID instead.

#### Cute Animal Picture

tiger-pig


pipermerriam:
Merging without review as this feels uncontroversial.  Happy to clean things up post merge if you have anything you'd like addressed.

lithp (commented):
sounds good :+1:

PR #131, opened 2020-10-22
Include cli args in Application initialization

pipermerriam:
## What was wrong?

When building up the CLI entry point for Alexandria I ended up needing access to the original `argparse.Namespace` object to be able to pull additional information off of the parser.

## How was it fixed?

Changed the `ApplicationAPI.__init__` signature to also include the parsed CLI args `argparse.Namespace` object.

I also changed the `Crawl` implementation to pull the concurrency value from CLI args rather than needing to change the initialization signature.

#### Cute Animal Picture

mouse-frog

### DIFF
--- a/ddht/abc.py
+++ b/ddht/abc.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+import argparse
 from collections import UserDict
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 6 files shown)


lithp (approved):
The premise sounds great, though I think the `ValueError` which checks that concurrency > 0 should happen somewhere during argument parsing. I've noticed code is always a lot simpler when it does parsing all in the same place!

PR #132, opened 2020-10-22
Register talk protocol in AlexandriaClient.__init__

pipermerriam:
## What was wrong?

The registering of the Alexandria `TalkProtocolAPI` with the `NetworkAPI` needs to happen reliably.

## How was it fixed?

Added logic to `AlexandriaClient.__init__` to register the protocol.

#### Cute Animal Picture

Unlikely-Animal-Friends-10


lithp (approved):
makes sense

PR #134, opened 2020-10-22
Fix race condition during shutdown of Network/Client/Dispatcher

### LINKED ISSUE #133: Flaky race condition
## What was wrong?

This has been occuring intermittently in the tests

```
DEBUG    async_service.Manager:trio.py:141 <Manager[Network] flags=SRCfe>: _handle_cancelled triggering task cancellation
DEBUG    async_service.Manager:trio.py:141 <Manager[Client] flags=SRCfe>: _handle_cancelled triggering task cancellation
DEBUG    async_service.Manager:trio.py:141 <Manager[Dispatcher] flags=SRCfe>: _handle_cancelled triggering task cancellation
ERROR    async_service.Manager:base.py:346 <Manager[Network] flags=SRCfe>: task _serve_find_nodes[daemon=True] exited with error:
Traceback (most recent call last):
  File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/async_service/base.py", line 324, in _run_and_manage_task
    await task.run()
  File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/async_service/trio.py", line 76, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 469, in _serve_find_nodes
    await self.client.send_found_nodes(
  File "/home/piper/projects/ddht/ddht/v5_1/client.py", line 244, in send_found_nodes
    await self.dispatcher.send_message(message)
  File "/home/piper/projects/ddht/ddht/v5_1/dispatcher.py", line 355, in send_message
    await self._outbound_message_send_channel.send(message)
  File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/trio/_channel.py", line 161, in send
    self.send_nowait(value)
  File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/trio/_core/_ki.py", line 160, in wrapper
    return fn(*args, **kwargs)
  File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/trio/_channel.py", line 141, in send_nowait
    raise trio.BrokenResourceError
trio.BrokenResourceError
```

Diagnosis isn't definitive, but I'm pretty sure what we're seeing here is a race condition in the long running background processes that is occuring during shutdown.

## How can it be fixed?

Not entirely sure.

pipermerriam:
fixes #133

## What was wrong?

We have long running processes in the `NetworkAPI` that interact with the `DispatcherAPI`.  During shutdown, the dispatcher is shut down first, which means that sometimes the network API tries to interact with the dispatcher API when it has already started closing channels.

## How was it fixed?

Somewhat hacky, but effective is to detect this within the external facing dispatcher API and if the error occurs **and** we are in the process of shutting down, we introduce a small sleep.  Since we're in the process of shutting down, in normal situations the sleep will be interrupted by a cancellation, and no error will occur.

#### Cute Animal Picture

12991778_768x1024

### DIFF
--- a/ddht/v5_1/dispatcher.py
+++ b/ddht/v5_1/dispatcher.py
@@ -352,7 +352,12 @@ class Dispatcher(Service, DispatcherAPI):
     async def send_message(self, message: AnyOutboundMessage) -> None:
         if message.receiver_node_id == self._pool.local_node_id:
             raise Exception("Cannot send message to self")
-        await self._outbound_message_send_channel.send(message)
+        try:
+            await self._outbound_message_send_channel.send(message)
+        except trio.BrokenResourceError:
+            if self.manager.is_cancelled:
+                await trio.sleep(1)
+            raise
 
     #
     # Request Response


lithp (approved):
Definitely a little hacky, but it works. If we end up needing to repeat this pattern we can pull it out into a context manager which does the exception conversion.

Looking to the long-term, it'd be even better if we didn't have to worry about this! Another way to get this effect might be to ensure that the network always shuts down before the dispatcher. Maybe a future version of async_service could allow you to declare your dependencies and it'll use that information to make sure that shutdown happens in the correct order.

pipermerriam:
> Maybe a future version of async_service could allow you to declare your dependencies and it'll use that information to make sure that shutdown happens in the correct order.

You can *almost* do this today.  If it is just tasks (and not child services) you can structure the scheduling of the tasks in the hierarchy you want them cancelled (so that dependent tasks are scheduled from within the tasks that they depend on).

```python
class MyService:
    async def run(self):
        self.manager.run_task(self._master_task)

    async def _master_task(self):
        send, receive = trio.open_memory_channel(0)
        self.manager.run_task(self._child_task, send)

        async with receive:
            async for thing in receive:
                 ...  # do stuff with the things

    async def _child_task(self, send):
        async with send:
            ...  # send things over the channel
```

In this example, we want to make sure we never try to send things over the channel when the receiving side has already been closed.  So we structure the tasks so that the receiving side is the parent of the task responsible for sending, ensuring that the task that sends will always be shut down before the task that receives.

But you can't really do this across service boundaries.  In this case, we'd need a way to push a task down into the task DAG across the service boundaries.  Interesting idea...

PR #136, opened 2020-10-22
Implement discv5_findNodes JSON-RPC API

pipermerriam:
## What was wrong?

The `discv5_findNodes` JSON-RPC endpoint is not implemented.

## How was it fixed?

Implemented an RPCHandler for it.

#### Cute Animal Picture

3429 custom

### DIFF
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -1,10 +1,17 @@
 import ipaddress
 from socket import inet_ntoa
-from typing import Any, Iterable, Optional, Tuple, TypedDict
+from typing import Any, Iterable, List, Optional, Tuple, TypedDict
 
 from eth_enr import ENR
 from eth_typing import HexStr, NodeID
-from eth_utils import decode_hex, is_hex, remove_0x_prefix, to_dict
+from eth_utils import (
+    decode_hex,
+    is_hex,
+    is_integer,
+    is_list_like,
+    remove_0x_prefix,
+    to_dict,
+)
 
 from ddht.abc import RPCHandlerAPI
 from ddht.endpoint import Endpoint
@@ -16,6 +23,28 @@ class PongResponse(TypedDict):
     enr_seq: int
     packet_ip: str
     packet_port: int
+
+
+def extract_params(request: RPCRequest) -> List[Any]:
+    try:
+        params = request["params"]
+    except KeyError:
+        raise RPCError("Request missing `params` key")
+
+    if not is_list_like(params):
+        raise RPCError(
+            f"Params must be list-like: params-type={type(params)} params={params}"
+        )
+
+    return params
+
+
+def validate_params_length(params: List[Any], length: int) -> None:
+    if len(params) != length:
+        raise RPCError(
+            f"Endpoint expects {length} params: length={len(params)} "
+            f"params={params}"
+        )
 
 
 def is_hex_node_id(value: Any) -> bool:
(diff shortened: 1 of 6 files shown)


tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -73,8 +100,99 @@ def ping_param_w3(request, alice, bob, bob_network):
...
+    enrs_at_0 = await trio.to_thread.run_sync(
+        w3.discv5.find_nodes, bob_node_id_param, 0
+    )
+    assert all(isinstance(enr, ENR) for enr in enrs_at_0)
+
+    # request with multiple distances
+    enrs_at_some_distance = await trio.to_thread.run_sync(
+        w3.discv5.find_nodes, bob_node_id_param, tuple(distances),
+    )
+    assert all(isinstance(ENR.from_repr(enr), ENR) for enr in enrs_at_some_distance)
```
lithp, line 198: `from_repr` will always return an `ENR`, I think the `isinstance` is extraneous.

I'm also surprised that both of these tests pass! If line 192 passes then `find_nodes` returns straight ENRs, if this line passes then `find_nodes` returns strings. Is it possible that one of these tests is not testing anything, because the result set is an empty list?

tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -73,8 +100,99 @@ def ping_param_w3(request, alice, bob, bob_network):
...
+    distances = set()
+
+    for _ in range(10):
+        enr = ENRFactory()
+        distances.add(compute_log_distance(bob.node_id, enr.node_id))
+        bob.enr_db.set_enr(enr)
+
+    # request with positional single distance
+    enrs_at_0 = await make_request("discv5_findNodes", [bob_node_id_param, 0])
+    assert all(isinstance(ENR.from_repr(enr), ENR) for enr in enrs_at_0)
```
lithp, line 137: Same thing on these lines, I think the `isinstance` is extraneous, and I think bucket 0 will always be empty. I think this test could pick an element for `distances`.

tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -33,29 +37,52 @@ async def bob_network(bob):
...
+        "unknown-endpoint",
+        "too-short",
+        "too-long",
+        "enode-missing-scheme",
+        "enode-missing-endpoint",
+        "enode-bad-nodeid",
+        "enr-without-prefix",
+    )
+)
+def invalid_node_id(request, alice, bob, bob_network):
```
lithp, line 64: this is a pretty neat way to get more understandable test failures!

ddht/v5_1/rpc_handlers.py
```diff
@@ -51,48 +80,47 @@ def validate_endpoint(value: Any) -> None:
...
 
-    def extract_params(self, request: RPCRequest) -> Tuple[NodeID, Optional[Endpoint]]:
-        try:
-            raw_params = request["params"]
-        except KeyError as err:
-            raise RPCError(f"Missiing call params: {err}")
+    if is_hex_node_id(value):
+        node_id = NodeID(decode_hex(value))
+        endpoint = None
+    elif value.startswith("enode://"):
```
lithp, line 90: I don't think there are any tests which cover this branch and then try to send a `FindNode` to the provided endpoint

ddht/v5_1/rpc_handlers.py
```diff
@@ -51,48 +80,47 @@ def validate_endpoint(value: Any) -> None:
...
 
-        value = raw_params[0]
+        raw_ip_address, _, raw_port = raw_endpoint.partition(":")
+        ip_address = ipaddress.ip_address(raw_ip_address)
+        port = int(raw_port)
+        endpoint = Endpoint(ip_address.packed, port)
+    elif value.startswith("enr:"):
+        enr = ENR.from_repr(value)
+        node_id = enr.node_id
+        endpoint = Endpoint.from_enr(enr)
```
lithp, line 105: If this raises an exception because the provided `enr` doesn't have enough information in it will that be convered into an `RPCError`?

lithp (approved):
I'm a little saddened that it takes 300 lines to add an RPC method and wire it up with some existing logic, but this PR looks good

ddht/v5_1/rpc_handlers.py
pipermerriam: It's covered under the parametrized fixture which enumerates the different supported formats (I verified by dropping an `assert False` into this context block and seeing the failure)

ddht/v5_1/rpc_handlers.py
pipermerriam: Good catch.  I'll add a test for this.

PR #138, opened 2020-10-22
Expand remaining AlexandriaNetworkAPI long running tasks

pipermerriam:
## What was wrong?

The alexandria network needs to maintain the routing table in the same manner as the base protocol.  Not all of that functionality has been ported over.

## How was it fixed?

Largely just a copy/pasta from `NetworkAPI` implementation with only superficial updates.

I explicitely didn't copy over any tests that are currently being applied to the equivalent functionality under the `NetworkAPI` because those tests are largely flakey/fragile and I don't want to expand that problem.

### DIFF
--- a/ddht/v5_1/alexandria/network.py
+++ b/ddht/v5_1/alexandria/network.py
@@ -1,15 +1,17 @@
 import itertools
 import logging
+import secrets
 from typing import Collection, List, Optional, Set, Tuple
 
 from async_service import Service
 from eth_enr import ENRAPI, ENRDatabaseAPI, ENRManagerAPI
 from eth_enr.exceptions import OldSequenceNumber
 from eth_typing import NodeID
-from eth_utils.toolz import take
+from eth_utils.toolz import cons, first, take
+from lru import LRU
 import trio
 
-from ddht._utils import reduce_enrs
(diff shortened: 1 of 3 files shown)


pipermerriam:
All of the code here is just copy/pasta from existing implementations in the `NetworkAPI` so review as you see fit.

ddht/v5_1/alexandria/network.py
```diff
@@ -248,6 +314,81 @@ async def _serve_find_nodes(self) -> None:
...
+
+                with trio.move_on_after(10):
+                    await self._routing_table_ready.wait()
+                    break
+
+        # TODO: Need better logic here for more quickly populating the
+        # routing table.  Should start off aggressively filling in the
+        # table, only backing off once the table contains some minimum
+        # number of records **or** searching for new records fails to find
+        # new nodes.  Maybe use a TokenBucket
```
lithp, line 381: By doing a couple random lookups we *should* be able to pretty quickly estimate of how large the network is, using similar logic to how a loglog counter works. From there we can know around how many nodes we expect a "full" routing table to have in each bucket, and start backing off once our distribution reaches that expected distribution.

lithp (approved):

Issue #115, opened 2020-10-13
Catch `trio.EndOfChannel` and re-raise as `trio.TooSlowError`


pipermerriam:
## What was wrong?

see: https://github.com/ethereum/ddht/pull/108#discussion_r503962915

The timeout mechanism used for the request/response API results in the outside call frame raising a `trio.EndOfChannel` exception when a timeout occurs.  This is not intuitive and will be confusing to anyone not intimately familiar with the internals.

See here for why the : https://github.com/ethereum/ddht/blob/21fcc6806ca49681c9d3d9c65dd0928c665e364d/ddht/v5_1/dispatcher.py#L441-L474

## How can it be fixed?

In call sites where this API is used, we should catch the `trio.EndOfChannel` and re-raise as a `trio.TooSlowError` so that it is obvious that we're seeing a timeout and not some obscure trio channel error.

PR #139, opened 2020-10-29
Catch EndOfChannel error and raise TooSlowError instead

njgheorghita:
## What was wrong?
Fixes https://github.com/ethereum/ddht/issues/115

## How was it fixed?

Summary of approach.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/v5_1/alexandria/test_alexandria_client.py
+++ b/tests/core/v5_1/alexandria/test_alexandria_client.py
@@ -131,6 +131,16 @@ async def test_alexandria_client_send_find_nodes(
         assert message.payload.distances == (0, 255, 254)
 
 
+@pytest.mark.trio
+async def test_alexandria_client_find_nodes_timeout(
+    bob, alice_alexandria_client, autojump_clock,
+):
+    with pytest.raises(trio.TooSlowError):
+        await alice_alexandria_client.find_nodes(
+            bob.node_id, bob.endpoint, distances=(0, 255, 254),
+        )
+
+
 @pytest.mark.parametrize(
     "enrs",
     (
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -215,7 +215,11 @@ class AlexandriaClient(Service, AlexandriaClientAPI):
                 )
                 try:
                     async with receive_channel:
-                        yield receive_channel
+                        try:
+                            yield receive_channel
+                        # Wrap EOC error with TSE to make the timeouts obvious
+                        except trio.EndOfChannel as err:
+                            raise trio.TooSlowError from err
                 finally:
                     nursery.cancel_scope.cancel()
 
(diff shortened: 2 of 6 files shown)


ddht/v5_1/client.py
```diff
@@ -405,7 +408,10 @@ async def talk(
             async with self.dispatcher.subscribe_request(
                 request, TalkResponseMessage
             ) as subscription:
-                return await subscription.receive()
+                try:
+                    return await subscription.receive()
+                except trio.EndOfChannel:
+                    raise trio.TooSlowError
```
pipermerriam, line 414:
```suggestion
                except trio.EndOfChannel as err:
                    raise trio.TooSlowError from err
```

pipermerriam (commented):
one more hiding here in the Alexandria part of the codebase: https://github.com/ethereum/ddht/blob/498f1e01cbff44d11d86321add30838a0a82b629/ddht/v5_1/alexandria/client.py#L217-L218

pipermerriam (approved):
Would you mind tossing in a copy/pasta of the timeout tests we use for the base-client into the alexandria tests so that we are excercising those code paths.  I've been running pretty fast in this codebase and tests aren't as exhaustive as I'd like.

pipermerriam:
(just a test for `AlexandriaClientAPI.ping()` timing out would be adequate.

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -131,6 +131,17 @@ async def test_alexandria_client_send_find_nodes(
         assert message.payload.distances == (0, 255, 254)
 
 
+@pytest.mark.trio
```
njgheorghita, line 134: @pipermerriam one final :eyes: for this test. This is the only test I could find that seems to test directly exercise the update to `AlexandriaClient.subscribe_request()`. From what I can tell, `AlexandriaClient` only uses the `subscribe_request()` api in `find_nodes()`.

pipermerriam: Correct!  didn't realize that the `AlexandriaClient.ping()` implementation goes directly through the `NetworkAPI.talk()`.  :+1:

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -131,6 +131,17 @@ async def test_alexandria_client_send_find_nodes(
...
+async def test_alexandria_client_send_find_nodes_timeout(
+    alice, bob, bob_client, alice_alexandria_client, autojump_clock,
+):
+    with pytest.raises(trio.TooSlowError):
+        async with bob_client.dispatcher.subscribe(TalkRequestMessage) as subscription:
```
pipermerriam, line 139: I think you can drop this line, as well as drop out the entire `bob_client` from this test.

pipermerriam (approved):

PR #141, opened 2020-10-29
Add LEB128 parser and encoder

pipermerriam:
## What was wrong?

We need tools for dealing with LEB128 encoded integers as part of the SSZ partial proofs used for content transmission in the Alexandria network.

## How was it fixed?

Copy/Pasta for the most part from the [implementation in py-wasm](https://github.com/ethereum/py-wasm/blob/41a6d07a620dfc4f590463dd038dffe4efe0c8c6/wasm/parsers/leb128.py)

#### Cute Animal Picture

motherhengoldenpup

### DIFF
--- a/ddht/exceptions.py
+++ b/ddht/exceptions.py
@@ -14,6 +14,14 @@ class DecodingError(BaseDDHTError):
     pass
 
 
+class ParseError(BaseDDHTError):
+    """
+    Raised as a generic error when trying to parse something.
+    """
+
+    pass
+
+
 class DecryptionError(BaseDDHTError):
     """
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
Looks good to me! Just wondering if it's useful to make the distinction that we're dealing with unsigned integers like in the `py-wasm` implementation?

pipermerriam:
> Looks good to me! Just wondering if it's useful to make the distinction that we're dealing with unsigned integers like in the py-wasm implementation?

I'm gonna leave as-is because the concise naming is nice and since it's an internal API we can rename if we end up needing a signed one as well.

PR #158, opened 2020-10-31
Faster bootstrapping process

pipermerriam:
## What was wrong?

The bootstrapping process was simple, leaving plenty of room for optimizations.

## How was it fixed?

First, small change to the initial bootstrapping logic to avoid an ineffective hang condition where bonding might happen after the `move_on_after` which would result in going back through the loop one more un-necessary time before moving on.

Second, I pulled the `TokenBucket` implementation from trinity so that we can burst our initial lookups, and then scale back to the periodic lookup when we are initially exploring the network.

Third, the actual lookup logic has been improved.  We still perform a random lookup, but we also perform a targeted lookup on one of our outermost non-full routing table buckets.  A randomly generated node-id has a 50% chance of landing in bucket 256.  This means that if we only perform random lookups, 50% of those will be for our outermost bucket.  Since the outermost bucket is typically always full, these lookups are not very effective.  By specifically targeted our non-full buckets we more effectively explore the unexplored parts of the network.

#### Cute Animal Picture

racoons-3-trash

### DIFF
--- a/ddht/_utils.py
+++ b/ddht/_utils.py
@@ -1,9 +1,13 @@
 import asyncio
+from contextlib import contextmanager
 import itertools
+import logging
+import math
 import operator
 import pathlib
 import secrets
 import socket
+import time
 from typing import (
     Any,
     AsyncGenerator,
@@ -12,7 +16,9 @@ from typing import (
     Callable,
     Collection,
     Iterable,
+    Iterator,
     Optional,
+    Sequence,
     Tuple,
     TypeVar,
     Union,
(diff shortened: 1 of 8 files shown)


pipermerriam:
I had to crank up some of the timeouts for the tests for recursive_find_nodes because now all of the 20 nodes we spin up do a bunch of network requests upon initialization.   I'm inclined to run with this and see if it becomes a problem at which point we can look at how to dumb the test down a bit.

marcgarreau (approved):

ddht/_utils.py
```diff
@@ -145,3 +151,49 @@ def reduce_enrs(enrs: Collection[ENRAPI]) -> Iterable[ENRAPI]:
...
+    finally:
+        end_at = time.monotonic()
+        elapsed = end_at - start_at
+        logging.getLogger("ddht.Timer").info("TIMER[%s]: %f", name, elapsed)
+
+
+TValue = TypeVar("TValue")
+
+
+def weighed_choice(values: Sequence[TValue]) -> TValue:
```
marcgarreau, line 170: was this intended to be `weighted`?

PR #145, opened 2020-10-30
Implement partial proofs

pipermerriam:
Builds on #175 

## What was wrong?

We need to be able to compute "partial" proofs.

## How was it fixed?

This PR fills out the `Proof.to_partial` object API and adds tests for it.

#### Cute Animal Picture

dogfox

PR #164, opened 2020-11-08
Implement ssz proof merging

pipermerriam:
Builds on #145 

## What was wrong?

When we request content in the Alexandria network, we get it in pieces.  We need a way to merge these pieces into a single unit.

## How was it fixed?

Quick and dirty implementation of `Proof.merge` which combines the two proofs into a single new proof, dropping any elements from the proof that are duplicated or no longer necessary (such as intermediate tree nodes for which we have all of the underlying leaves.

#### Cute Animal Picture

20-Absolutely-Amazing-Dog-Halloween-Costumes-4

### DIFF
--- a/ddht/v5_1/alexandria/partials/_utils.py
+++ b/ddht/v5_1/alexandria/partials/_utils.py
@@ -1,6 +1,7 @@
 from typing import Iterable
 
 from eth_utils import to_tuple
+from eth_utils.toolz import sliding_window
 from ssz.constants import CHUNK_SIZE
 
 from ddht.v5_1.alexandria.constants import POWERS_OF_TWO
@@ -47,3 +48,22 @@ def get_chunk_count_for_data_length(length: int) -> int:
     if length == 0:
         return 0
     return (length + CHUNK_SIZE - 1) // CHUNK_SIZE  # type: ignore
+
+
+@to_tuple
+def filter_overlapping_paths(*paths: TreePath) -> Iterable[TreePath]:
+    """
+    Filter out any paths that are a prefix of another path.
+    """
+    if not paths:
+        return
+    sorted_paths = sorted(paths)
+    for left, right in sliding_window(2, sorted_paths):
+        if right[: len(left)] == left:
+            continue
+        else:
+            yield left
+
+    # Because of the use of `sliding_window` we need to manually yield the last
+    # path
+    yield sorted_paths[-1]
--- a/ddht/v5_1/alexandria/partials/proof.py
+++ b/ddht/v5_1/alexandria/partials/proof.py
@@ -25,6 +25,7 @@ from ssz.sedes import List as ListSedes
 from ddht.v5_1.alexandria.partials._utils import (
     decompose_into_powers_of_two,
     display_path,
+    filter_overlapping_paths,
     get_chunk_count_for_data_length,
 )
 from ddht.v5_1.alexandria.partials.chunking import (
@@ -108,13 +109,11 @@ class DataPartial:
         else:
             raise TypeError(f"Unsupported type: {type(index_or_slice)}")
 
-        data_length = end_at - start_at
-
         candidate_index = max(0, bisect.bisect_left(self._segments, (start_at,)) - 1)
         segment = self._segments[candidate_index]
         is_within_bounds = (
             segment.start_index <= start_at <= segment.end_index
-            and data_length <= len(segment.data)
+            and end_at <= segment.end_index
         )
         if not is_within_bounds:
             raise IndexError(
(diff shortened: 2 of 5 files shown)


ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -344,6 +345,45 @@ def get_proven_data_segments(self, length: int) -> Iterable[Tuple[int, bytes]]:
...
+            if path in all_other_elements_by_path
+        )
+
+        all_merged_elements = (
+            common_elements + merged_elements_from_self + merged_elements_from_other
+        )
+
+        proof = Proof(self.hash_tree_root, all_merged_elements, self.sedes)
+        validate_proof(proof)
+        return proof
```
marcgarreau, line 385: Context gathering: So this merge applies to two partial proofs. What is the upper bounds of the number of partial proofs required for some data request? Is there some workflow for more than two proofs or is that a recursive use of this functionality, or something else?

pipermerriam: Data requests in alexandria end up transferring about 750 bytes of data (and about 400 bytes of overhead from the merkle stuff).  According to Etherscan and some very quick sampling, ethereum blocks are about 45kb which means it'll take about 60 messages to transfer all the data, which means 60 partial proofs that end up getting re-assembled.  You can see the preliminary code that does this here: 

https://github.com/ethereum/ddht/pull/165/files#diff-e8d0d620a600d82775b4794f8378624f90f1a7c0e9ed3e692da2a39e2a10de89R286-R480

pipermerriam: You can use this to extrapolate some of the other data payloads.  Headers will always be sent in a single message because they are not big enough to need to split them up.  Individual transactions will typically fit in a single message (and thus won't require a proof) but sometimes will need to be split up.  etc, etc.

ddht/v5_1/alexandria/partials/_utils.py
```diff
@@ -47,3 +48,15 @@ def get_chunk_count_for_data_length(length: int) -> int:
...
+@to_tuple
+def merge_paths(*paths: TreePath) -> Iterable[TreePath]:
+    sorted_paths = sorted(paths)
+    for left, right in sliding_window(2, sorted_paths):
+        if right[: len(left)] == left:
+            continue
+        else:
+            yield left
+
+    yield from sorted_paths[-1:]
```
lithp, line 62: Why is this not `yield sorted_paths[-1]`?

ddht/v5_1/alexandria/partials/_utils.py
```diff
@@ -47,3 +48,15 @@ def get_chunk_count_for_data_length(length: int) -> int:
     if length == 0:
         return 0
     return (length + CHUNK_SIZE - 1) // CHUNK_SIZE  # type: ignore
+
+
+@to_tuple
+def merge_paths(*paths: TreePath) -> Iterable[TreePath]:
```
lithp, line 54: A docstring could help here, I'm not sure what this function is doing. It takes a bunch of paths and de-duplicates them? If any path is a prefix of another the shorter path is dropped. From the name it sounded like it would somehow combine paths to make other paths, if I've got what it's doing right maybe this could be called `deduplicate_paths`.

ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -29,6 +36,84 @@ def __str__(self) -> str:
...
+    start_index: int, num_padding_chunks: int, path_bit_length: int
+) -> Iterable[ProofElement]:
+    """
+    Get the padding elements for a proof.
+
+    By decomposing the number of chunks needed into the powers of two which
+    make up the number we can construct the minimal right hand sub-tree(s)
+    needed to pad the hash tree.
+    """
+    for power_of_two in decompose_into_powers_of_two(num_padding_chunks):
```
lithp, line 50: I'm not sure this name is any better, but it took me a second to figure out what this method was doing. It might be more clear if this method was called `nonzero_bits`?

lithp: Also, this is black magic! The docstring helps me understand what is happening but I think I might need a picture or a longer docstring to understand why this works.

ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -29,6 +36,84 @@ def __str__(self) -> str:
...
+    def __init__(self, length: int, segments: Tuple[Tuple[int, bytes], ...]) -> None:
+        self._length = length
+        self._segments = segments
+
+    def __len__(self) -> int:
+        return self._length
+
+    @overload
+    def __getitem__(self, index_or_slice: int) -> int:
+        ...
```
lithp, line 75: Python must be so strange for anybody learning it from scratch today :sweat_smile:

ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -29,6 +36,84 @@ def __str__(self) -> str:
...
+            end_at = index_or_slice.stop
+        elif isinstance(index_or_slice, int):
+            start_at = index_or_slice
+            end_at = index_or_slice + 1
+        else:
+            raise TypeError(f"Unsupported type: {type(index_or_slice)}")
+
+        data_length = end_at - start_at
+
+        candidate_index = max(0, bisect.bisect_left(self._segments, (start_at,)) - 1)
```
lithp, line 95: Probably worth putting into the docstring that this class assumes `segments` is sorted, or sorting in the constructor. If I remember right Python's timsort has enough special cases that already-sorted lists are sorted pretty quick.

ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -29,6 +36,84 @@ def __str__(self) -> str:
...
+        yield ProofElement(left_path[: path_bit_length - depth], padding_hash_tree_root)
+
+
+class DataPartial:
+    """
+    A wrapper around a partial data proof which allows data retrieval by
+    indexing or slicing.
+
+    Raise `IndexError` if the requested data is not part of the proof.
+    """
```
lithp, line 64: Even after reading the tests I'm still not sure what this class is for. You pass in a bunch of (path, data) tuples, and then you can read them back out by searching for them by path or path range?

ddht/v5_1/alexandria/partials/_utils.py
pipermerriam: I think it's something like `remove_overlapped_paths` as it removes any path that is on another longer path.

ddht/v5_1/alexandria/partials/_utils.py
pipermerriam: no clue, but they should be equivalent and you're is more sane.  will fix.

ddht/v5_1/alexandria/partials/proof.py
pipermerriam: Yeah, I can make another pretty picture to help make it clear how/why this works.  It is indeed black magic.  Building up the intuition about how all of these different tree mechanics work was hard.  Good to write it down while I still remember it all.

ddht/v5_1/alexandria/partials/proof.py
pipermerriam: Yeah, a `Proof` houses *some* amount of the actual data (maybe all of it, maybe a subsection).  This abstracts that away so that you can just treat it like it's a bytestring and directly access the data via slice/index mechanics.  I'll beef up the docstring a bit.

ddht/v5_1/alexandria/partials/proof.py
marcgarreau: ^ lol, not too bad in my experience. As someone not familiar with all the conventions: are we using `...` to mean `TODO` or `noop`?

pipermerriam:
This has been rebased.  @marcgarreau and @lithp do either of you have anything additional you'd like me to cleanup?

Issue #161, opened 2020-11-04
Improve bond(...) API logic to handle missing endpoint information


pipermerriam:
## What was wrong?

Currently, the two `bond(...)` APIs found on `NetworkAPI` and `AlexandriaNetworkAPI` fail if we do not have an `ENR` record for the `node_id` we are trying to bond with.

Similarly, if our ENR record is outdated, the bond will fail.  We can potentially recover from this case via the same mechanism, potentially updating our ENR record and retrying if the search results in updated Endpoint information in the ENR record.

## How can it be fixed?

We should be able to improve the `bond(...)` logic such that *if* the `Endpoint` is unknown it does a `recursive_find_node` in an attempt to discover an ENR record for the `node_id` in question.

This is especially important in Alexandria where advertisements are linked to a `node_id` but do not carry information about the node's endpoint.  Part of the validation logic for advertisements is a liveliness check which will fail in cases where we don't yet have an ENR record for them.

pipermerriam:
Logic should look roughly like:

- A: If endpoint information is provided, try to bond and use current logic.
- B: If endpoint information is not explicitely provided, look it up from database.
    - B.1 If there **is not** endpoint in the database: do a `recursive_find_nodes` against the `node_id`
    - B.2 If the endpoint information **is** in the database: but the `ping` times out, perform a `recursive_find_nodes` against the `node_id`
    - B.3 If the `recursive_find_nodes` from B.1 or B.2 returns an ENR record for the node **and** the ENR record is newer than whatever we had on hand (or we if we didn't have one), try the `ping` again with the new endpoint information from the ENR record.

We only fail under the conditions:

1. explicitly provide `endpoint` and ping fails
2. ping fails and we cannot get an ENR record via `recursive_find_nodes`
3. `recursive_find_nodes` yields a new ENR record but the `ping` fails anyways.

PR #166, opened 2020-11-09
Refactor lazy endpoint retrieval for nodes

pipermerriam:
Fixes https://github.com/ethereum/ddht/issues/161

## What was wrong?

Up until now, the only way for us to learn about another node was via `FINDNODES/FOUNDNODES`.  This had the implicit guarantee that we would have an ENR record for that node.

With the upcoming introduction of Advertisements in Alexandria, we gain a new way to find a `node_id` without having encountered the node's ENR record yet.  This makes things like performing the liveliness check difficult.

## How was it fixed?

First, I removed some of the duplicated `AlexandriaNetworkAPI` methods.  Specifically, it now delegates to `NetworkAPI.lookup_enr` and `NetworkAPI.endpoint_for_node_id`.

Second, the `NetworkAPI.lookup_enr` now falls back to trying to find the node via a `NetworkAPI.recursive_find_nodes` when the endpoint is not known.

This also implements a new API, `NetworkAPI.get_nodes_near` which exposes a combined API between `NetworkAPI.recursive_find_nodes` and `RoutingTableAPI.iter_nodes_around` allowing for a more robust way to reliably find the closest nodes we can to a certain area of the network.

#### Cute Animal Picture

redpanda

### DIFF
--- a/ddht/v5_1/abc.py
+++ b/ddht/v5_1/abc.py
@@ -505,3 +505,11 @@ class NetworkAPI(ServiceAPI):
     @abstractmethod
     async def recursive_find_nodes(self, target: NodeID) -> Tuple[ENRAPI, ...]:
         ...
+
+    @abstractmethod
+    async def get_nodes_near(self, target: NodeID) -> Tuple[NodeID, ...]:
+        ...
+
+    @abstractmethod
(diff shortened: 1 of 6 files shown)


pipermerriam:
need to add tests for this new functionality.

lithp (approved):
looks great

PR #167, opened 2020-11-09
Expose local private key from ClientAPI

pipermerriam:
## What was wrong?

The Alexandria network needs access to the private key in order to sign advertisements.

## How was it fixed?

Expose it as a public property.

#### Cute Animal Picture

dogs-carrying-piano-pet-costume


pipermerriam:
Merging as this shouldn't be controversial.  Lemme know if I'm wrong

PR #168, opened 2020-11-09
More precice upper bounds on alexandria payloads

pipermerriam:
## What was wrong?

We should make sure we every byte we can in our UDP packets.

## How was it fixed?

A more precise calculation of the upper bound we can expect for packets.

#### Cute Animal Picture

Insta-Barney-mixedbreed-TheBilionaire-scasserly1

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -29,7 +29,7 @@ from ddht.request_tracker import RequestTracker
 from ddht.subscription_manager import SubscriptionManager
 from ddht.v5_1.abc import NetworkAPI
 from ddht.v5_1.alexandria.abc import AlexandriaClientAPI
-from ddht.v5_1.alexandria.constants import ALEXANDRIA_PROTOCOL_ID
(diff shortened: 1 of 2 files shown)


ddht/v5_1/alexandria/client.py
```diff
@@ -141,7 +142,10 @@ async def _request(
         # then feed this into our local tracker, which allows us to query it
         # upon receiving an incoming TALKRESPONSE to see if the response is to
         # a message from this protocol.
-        request_id = self.network.client.request_tracker.get_free_request_id(node_id)
+        if request_id is None:
```
pipermerriam, line 145: this should get pulled out.

lithp (approved):
It'd be cool if `MAX_PAYLOAD_SIZE` was the result of some kind of computation, so that we don't need to remember to change it when the header format changes. Looks good though :+1:

pipermerriam:
I thought about computing it but I decided against it because:  1) if the bounds are too high for some reason our tests will actually fail on trying to send the datagram packets. 2) if the bounds are too low... well, then everything will work fine, just slightly less efficiently.

I figure we can come back later and squeeze out a few extra bytes if we really want to, but this is pretty close to optimal.

PR #170, opened 2020-11-09
Add ResourceQueue

pipermerriam:
## What was wrong?

When retrieving content from the Alexandria network, we'll have the following two sets of things that need special handling.

- A list of nodes that have the content.
- A list of "chunks" that we still need.

For both of these we need a way to manage ensuring that we aren't hitting the same node with multiple requests for content as well as making sure we aren't requesting the same chunk from multiple peers.

## How was it fixed?

This implements a convenience layer over what *could* be done more verbosely with raw trio Channels.

```python
# with raw channels
resources = (...,)
send, receive = trio.open_memory_channel(...)
for resource in resources:
    send.send_nowait(resource)

def worker(...):
    resource = await receive.receive()
    try:
        ...  # do something with it
    except:
        await send.send(resource)  # put it back in the queue if there's an error
    else:
        if not done_with_resource:
            await send.send(resource)

# With the Resource Queue
resources = (...,)
queue = ResourceQueue(resources)

def worker(...):
    async with queue.reserve() as resource:
        ...  # do something with it
        if done_with_resource:
            queue.remove(resource)
```

This pattern cleans the code up a lot, especially in the case where we have two sets of resources that are being concurrently managed.

#### Cute Animal Picture

20-Absolutely-Amazing-Dog-Halloween-Costumes-1

### DIFF
--- a/ddht/v5_1/alexandria/resource_queue.py
+++ b/ddht/v5_1/alexandria/resource_queue.py
@@ -11,11 +11,47 @@ class RemoveResource(Exception):
 
 
 class ResourceQueue(Container[TResource]):
+    """
+    Allow some set of "worker" processes to share the underlying resources,
+    ensuring that any given resource is only in use by a single worker at any
+    given time.
+
+    .. code-block:: python
+
+        things = (thing_1, thing_2, ...)
+        queue = ResourceQueue(things)
+
+        async def worker():
+            while True:
+                # within the following async context block the `resource` is
+                # reserved for this worker.
+                async with queue.reserve() as resource:
+
+                    # We typically want to do something with the resource.
+                    do_work(resource)  # do
+
+                    # We can remove the resource from the queue
+                    if done_with(resource):
+                        queue.remove()
+
+                    # We can add new resources to the queue
+                    if has_new_resources(resource):
+                        new_resource = ...  #
+                        queue.add(new_resource)
+
+                # Upon exiting the context block the resource is automatically
+                # added back into the queue (unless it was explicitely
+                # removed).
+    """
+
     resources: Set[TResource]
 
     def __init__(
-        self, resources: Collection[TResource], max_resource_count: int = 256
+        self, resources: Collection[TResource], max_resource_count: int = None,
     ) -> None:
+        if max_resource_count is None:
+            max_resource_count = len(resources)
+
         if len(resources) > max_resource_count:
             raise ValueError(
                 f"Number of resources exceeds maximum: {len(resources)} > {max_resource_count}"
@@ -42,6 +78,9 @@ class ResourceQueue(Container[TResource]):
 
     @asynccontextmanager
     async def reserve(self) -> AsyncIterator[TResource]:
+        # Fetch a new resource from the queue.  If the resource is no longer
+        # part of the tracked resources discard it and move onto the next
+        # resource in the queue.
         while True:
             resource = await self._receive.receive()
             if resource in self:
@@ -52,5 +91,8 @@ class ResourceQueue(Container[TResource]):
         try:
             yield resource
         finally:
+            # The resource could have been removed during the context block so
+            # only add it back to the queue if it is still part of the tracked
+            # resources.
             if resource in self:
                 await self._send.send(resource)
--- a/ddht/v5_1/alexandria/resource_queue.py
+++ b/ddht/v5_1/alexandria/resource_queue.py
@@ -1,0 +1,56 @@
+from typing import Any, AsyncIterator, Collection, Container, Hashable, Set, TypeVar
+
+from async_generator import asynccontextmanager
+import trio
+
+TResource = TypeVar("TResource", bound=Hashable)
+
+
+class RemoveResource(Exception):
+    pass
+
+
+class ResourceQueue(Container[TResource]):
+    resources: Set[TResource]
+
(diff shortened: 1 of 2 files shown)


ddht/v5_1/alexandria/resource_queue.py
```diff
@@ -0,0 +1,56 @@
...
+        self.resources.remove(resource)
+
+    @asynccontextmanager
+    async def reserve(self) -> AsyncIterator[TResource]:
+        while True:
+            resource = await self._receive.receive()
+            if resource in self:
+                break
+            else:
+                continue
```
lithp, line 50: When will the resource not be `in self`?

tests/core/v5_1/alexandria/test_resource_queue.py
```diff
@@ -0,0 +1,77 @@
...
+    for _ in range(random.randint(0, num) + base):
+        await trio.lowlevel.checkpoint()
+
+
+@pytest.mark.trio
+async def test_resource_queue():
+    known_resources = {"a", "b", "c", "d"}
+    queue = ResourceQueue(known_resources)
+
+    resources_in_use = set()
```
lithp, line 19: I'm not sure what `resources_in_use` is meant to be testing. If I'm reading this right then there is no implementation of `ResourceQueue` which cause the `resources_in_use` assertions to fail.

ddht/v5_1/alexandria/resource_queue.py
pipermerriam: If `queue.remove()` is called within the context block.

```python
async with queue.reserve() as resource:
    ...  # do something with it
    if some_condition:
        queue.remove(resource)
```

tests/core/v5_1/alexandria/test_resource_queue.py
pipermerriam: Yeah, it's a dirty way for me to test the "property" that any resource that is in use should never be allocated in a call to `queue.reserve`.  I recognize this is effectively a re-implementation of the business logic using stateful local variables.  I'm open to an alternate approach if you've got a suggestion.

ddht/v5_1/alexandria/resource_queue.py
```diff
@@ -0,0 +1,56 @@
...
+class RemoveResource(Exception):
+    pass
+
+
+class ResourceQueue(Container[TResource]):
+    resources: Set[TResource]
+
+    def __init__(
+        self, resources: Collection[TResource], max_resource_count: int = 256
+    ) -> None:
```
lithp, line 18: I think there's a bug if `resources` has more than 256 elements and the user does not provide a `max_resource_count`, if nothing is passed in it might want to default to something like `len(resources) + 256`.

This interface would be much easier to implement if we could resize memory channels. It looks like we *can*, by changing `self._send._state.max_buffer_size`, but I don't know how you feel about hard-coding trio implementation details.

I don't know where you plan on using this class so I don't know if this suggestion makes sense, but the easiest way to remove this class of bugs is to not have an `add(resource)` method.

ddht/v5_1/alexandria/resource_queue.py
```diff
@@ -0,0 +1,56 @@
...
+import trio
+
+TResource = TypeVar("TResource", bound=Hashable)
+
+
+class RemoveResource(Exception):
+    pass
+
+
+class ResourceQueue(Container[TResource]):
```
lithp, line 13: Future readers will appreciate a docstring here, it took me a few minutes to figure out what this class was going! This is kind of like a [`CapacityLimiter`](https://trio.readthedocs.io/en/v0.12.1/reference-core.html#trio.CapacityLimiter) except each token has a name?

ddht/v5_1/alexandria/resource_queue.py
lithp: Ahh, good point

tests/core/v5_1/alexandria/test_resource_queue.py
lithp: Ahh, sorry, I missed that there are 10 instances of worker, this makes sense now.

I bet some combination of [hypothesis](https://hypothesis.readthedocs.io/en/latest/stateful.html) or [sequencer](https://trio.readthedocs.io/en/v0.12.1/reference-testing.html#trio.testing.Sequencer) could help make this test simpler, but this already works so I definitely won't hold the review over it.

lithp (approved):
left some comments but after those this looks good

ddht/v5_1/alexandria/resource_queue.py
pipermerriam: It's used here: 

https://github.com/ethereum/ddht/pull/95/files#diff-e8d0d620a600d82775b4794f8378624f90f1a7c0e9ed3e692da2a39e2a10de89R432-R503

The `add(resource)` is needed so that when we request a section of the content, and the response doesn't contain all of the requested data, we can add the sub-sections that are still missing back into the queue.

For now I'm not really intending to use this beyond that location.  I'll figure out a patch for the bug.

ddht/v5_1/alexandria/resource_queue.py
pipermerriam: More like each token is an actual "thing" but yes. similar constructs.

PR #171, opened 2020-11-09
Extend proof api checking proof completion

pipermerriam:
Builds on #164 

## What was wrong?

When retrieving content over Alexandria that is larger than a single packet, we transmit the content using SSZ partials.  We need a few more APIs on the proof object to facilitate this.

- What parts of the data are missing
- Is the proof complete
- Does the proof contain a specific chunk

## How was it fixed?

I added the following APIs

- `Proof.get_missing_segments()` returns the parts of the data that are missing.
- `Proof.is_complete()` returns whether the proof contains all of the data chunks.
- `Proof.has_chunk(...)` return whether a specific chunk index is in the proof.

#### Cute Animal Picture

cute-red-panda-13

### DIFF
--- a/ddht/v5_1/alexandria/partials/chunking.py
+++ b/ddht/v5_1/alexandria/partials/chunking.py
@@ -1,5 +1,6 @@
+import bisect
 import itertools
-from typing import Iterable, List, Sequence, Tuple
+from typing import Iterable, List, NamedTuple, Sequence, Tuple
 
 from eth_typing import Hash32
 from eth_utils import to_tuple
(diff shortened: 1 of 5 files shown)


ddht/v5_1/alexandria/partials/proof.py
```diff
@@ -468,7 +482,7 @@ def to_partial(self, start_at: int, partial_data_length: int) -> "Proof":
         #
         # start-at : last_partial_chunk_index + 1
         # end-at   : N/A
-        if last_data_chunk_index + 1 <= self.sedes.chunk_count - 1:
+        if last_data_chunk_index + 1 <= self.sedes.chunk_count:
```
pipermerriam, line 485: I need to unwind these bugfixes from the new features.

pipermerriam:
@marcgarreau or @lithp if you would like to just look at the last commit (currently 3a834ad) you can view a more minimized diff. I did a small bit of code-organization in the first commit that makes the diff larger than necessary.

lithp (approved):
looks good to me :+1:

ddht/v5_1/alexandria/partials/chunking.py
```diff
@@ -222,3 +223,51 @@ def group_by_subtree(
...
+        """
+        insertion_indices = bisect.bisect(
+            tuple(other.end_at for other in segments), self.start_at
+        )
+        for other in segments[insertion_indices:]:
+            if other.start_at >= self.end_at or other.end_at <= self.start_at:
+                break
+            start_at = max(self.start_at, other.start_at)
+            end_at = min(self.end_at, other.end_at)
+            yield MissingSegment(start_at, end_at - start_at)
```
marcgarreau, line 253: clever implementation 🚀

marcgarreau (approved):

PR #173, opened 2020-11-10
Add discv5_sendPing json rpc endpoint

njgheorghita:
## What was wrong?
Support the `discv5_sendPing` jsonrpc endpoint.

## How was it fixed?
Add a jsonrpc endpoint for `discv5_sendPing` that will send a ping to the given node id, but won't wait for a pong in response like `discv5_ping` and will simply return the request id.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -6,12 +6,12 @@ from eth_enr import ENR
 from eth_typing import HexStr, NodeID
 from eth_utils import (
     decode_hex,
+    encode_hex,
     is_hex,
     is_integer,
     is_list_like,
     remove_0x_prefix,
     to_dict,
-    to_int,
 )
 
 from ddht.abc import RPCHandlerAPI
@@ -27,7 +27,7 @@ class PongResponse(TypedDict):
 
 
 class SendPingResponse(TypedDict):
-    request_id: int
+    request_id: HexStr
 
 
 def extract_params(request: RPCRequest) -> List[Any]:
@@ -158,7 +158,7 @@ class SendPingHandler(RPCHandler[Tuple[NodeID, Optional[Endpoint]], SendPingResp
             enr = await self._network.lookup_enr(node_id)
             endpoint = Endpoint.from_enr(enr)
         request_id = await self._network.client.send_ping(node_id, endpoint)
-        return SendPingResponse(request_id=to_int(request_id))
+        return SendPingResponse(request_id=encode_hex(request_id))
 
 
 def _is_valid_distance(value: Any) -> bool:
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -11,6 +11,7 @@ from eth_utils import (
     is_list_like,
     remove_0x_prefix,
     to_dict,
+    to_int,
 )
 
 from ddht.abc import RPCHandlerAPI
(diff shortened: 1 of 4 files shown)


tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -129,6 +129,28 @@ async def test_v51_rpc_ping_web3(make_request, bob_node_id_param_w3, alice, bob,
...
+    response = await make_request("discv5_sendPing", [bob_node_id_param])
+    assert response["request_id"] > 0
+
+
+@pytest.mark.trio
+async def test_v51_rpc_send_ping_invalid_node_id(make_request, invalid_node_id):
+    with pytest.raises(Exception, match="'error':"):
+        await make_request("discv5_sendPing", [invalid_node_id])
+    with pytest.raises(Exception, match="'error':"):
+        await make_request("discv5_sendPing", [])
```
pipermerriam, line 143: Since `invalid_node_id` is a parametrized fixture, I think that this test will get run multiple times.  I would recommend shifting it out into it's own test.

ddht/v5_1/rpc_handlers.py
```diff
@@ -25,6 +26,10 @@ class PongResponse(TypedDict):
     packet_port: int
 
 
+class SendPingResponse(TypedDict):
+    request_id: int
```
pipermerriam, line 30: Request ID should be treated as `bytes` in all v51 contexts.

pipermerriam (approved):

ddht/v5_1/rpc_handlers.py
```diff
@@ -25,6 +26,10 @@ class PongResponse(TypedDict):
     packet_port: int
 
 
+class SendPingResponse(TypedDict):
+    request_id: HexStr
```
njgheorghita, line 30: > Request ID should be treated as `bytes` in all v51 contexts.

@pipermerriam Is it incorrect to use `HexStr` here? Using raw `bytes` causes problems with the json serialization.

tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -129,6 +134,36 @@ async def test_v51_rpc_ping_web3(make_request, bob_node_id_param_w3, alice, bob,
     assert pong.packet_port == alice.endpoint.port
 
 
+@pytest.mark.trio
+async def test_v51_rpc_send_ping(make_request, bob_client, bob):
```
njgheorghita, line 138: @pipermerriam When using the `bob_node_id_param` fixture here (as well as the `bob_node_id_param_w3` fixture for the `test_v51_rpc_send_ping_web3` test) it causes a strange error that I'm unfamiliar with: `trio.WouldBlock: Exception: Locked: owner=Node.network(...)  acquiring-for=Node.client(...)`. 

This error goes away when using `repr(bob.enr)` as the node id - but it seems to me that we would want to test all the node id flavors in these tests. Do you have an idea as to what's causing this failure?

ddht/v5_1/rpc_handlers.py
pipermerriam: Totally fine to return it as a hex encoded value. :+1:

tests/core/v5_1/test_v51_rpc_handlers.py
pipermerriam: Ah, ok, so that's some safety code I wrote to ensure that you don't use `bob_network` and `bob_client` at the same time.  The reason is that `bob_network` runs the client under the hood so you'll run into an error trying to bind to the socket because the port will already be used.

So if you need both `bob_client` and `bob_network` within the same test, you should use `bob_network` and pull the client directly off of it as `bob_network.client`

PR #176, opened 2020-11-12
Improve NetworkAPI.get_nodes_near

pipermerriam:
## What was wrong?

The `get_nodes_near` API was still lacking (and still sort of is).  It didn't try very hard to query the targeted part of the network.

## How was it fixed?
Since the nodes closest to the target have the most information about that part of the network, the algorithm now finds the closest nodes we can, and then does additional `FINDNODES` requests to those nodes to fill out a more complete picture of the target.  

> Note: The current implementation isn't intended to be final.  There's a number of efficiency improvements that will end up being necessary to deal with unresponsive nodes, but I'm kicking that can down the road for now.

#### Cute Animal Picture

red-panda-2

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -39,6 +39,19 @@ from ddht.v5_1.messages import (
     TalkRequestMessage,
 )
 
+NEIGHBORHOOD_DISTANCES = (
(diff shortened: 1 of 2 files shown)


lithp (approved):
looks good to me!

PR #178, opened 2020-11-13
Implement enr db management jsonrpc endpoints.

njgheorghita:
## What was wrong?
Add jsonrpc support for `discv5_setENR`, `discv5_getENR`, `discv5_deleteENR` endpoints.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -3,6 +3,7 @@ from typing import Any, Iterable, List, Optional, Tuple, TypedDict
 
 from eth_enr import ENR
 from eth_enr.abc import ENRAPI
+from eth_enr.exceptions import OldSequenceNumber
 from eth_typing import HexStr, NodeID
 from eth_utils import ValidationError, decode_hex, encode_hex, is_list_like, to_dict
 
@@ -172,7 +173,10 @@ class SetENRHandler(RPCHandler[ENRAPI, None]):
         return enr
 
     async def do_call(self, params: ENRAPI) -> None:
-        self._network.enr_db.set_enr(params)
+        try:
+            self._network.enr_db.set_enr(params)
+        except OldSequenceNumber as exc:
+            raise RPCError(f"Invalid ENR, outdated sequence number: {exc}.")
         return None
 
 
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -1,8 +1,10 @@
 from socket import inet_ntoa
 from typing import Any, Iterable, List, Optional, Tuple, TypedDict
 
+from eth_enr import ENR
+from eth_enr.abc import ENRAPI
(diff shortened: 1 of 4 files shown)


ddht/v5_1/rpc_handlers.py
```diff
@@ -115,8 +121,62 @@ async def do_call(self, params: FindNodesRPCParams) -> Tuple[str, ...]:
...
+        validate_params_length(raw_params, 1)
+        enr_repr = raw_params[0]
+        try:
+            enr = ENR.from_repr(enr_repr)
+        except ValidationError:
+            raise RPCError(f"Invalid ENR repr: {enr_repr}")
+        return enr
+
+    async def do_call(self, params: ENRAPI) -> None:
+        self._network.enr_db.set_enr(params)
```
pipermerriam, line 155: Unhandled error here if the enr_db already has this record in the DB.  I think you can trigger this in tests by just sending in the same ENR twice.

tests/core/v5_1/test_v51_rpc_handlers.py
```diff
@@ -135,24 +153,86 @@ async def test_v51_rpc_ping_web3(make_request, bob_node_id_param_w3, alice, bob,
...
+async def test_v51_rpc_get_enr_with_unseen_node_id(make_request, new_enr):
     with pytest.raises(Exception, match="'error':"):
-        await make_request("discv5_sendPing", [invalid_node_id])
+        await make_request("discv5_getENR", [repr(new_enr)])
 
 
 @pytest.mark.trio
-async def test_v51_rpc_send_ping_missing_node_id(make_request):
+async def test_v51_rpc_get_enr_web3(make_request, bob, bob_node_id_param_w3, w3):
+    response = await trio.to_thread.run_sync(w3.discv5.get_enr, bob_node_id_param_w3)
```
pipermerriam, line 169: What happens if the ENR isn't in the database?  (I think that the `w3` API should raise an exception in this case similar to what we did with `w3.eth.getTransaction` a while back so that we didn't have to do return value checking.

pipermerriam (approved):

ddht/v5_1/rpc_handlers.py
njgheorghita: It looks to me that `enr_db` will only throw an error if the sequence number of the new enr is older than the existing enr [link](https://github.com/ethereum/eth-enr/blob/a9268bc16bf54881ee1e0c5245bb95f8aabe0176/eth_enr/enr_db.py#L49), so sending in the same enr wouldn't trigger this exception, though I could be missing something. I've added a test to catch the case with an outdated sequence number.

tests/core/v5_1/test_v51_rpc_handlers.py
njgheorghita: Yup, it'll raise an exception, I've added a test case to show this.

PR #182, opened 2020-11-17
Add support for discv5_updateNodeInfo endpoint

njgheorghita:
## What was wrong?
Add jsonrpc support for `discv5_updateNodeInfo`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/rpc_handlers.py
+++ b/ddht/rpc_handlers.py
@@ -4,7 +4,15 @@ from eth_enr import ENRAPI
 from eth_enr.abc import ENRManagerAPI
 from eth_enr.typing import ENR_KV
 from eth_typing import HexStr
-from eth_utils import encode_hex, is_integer, is_text, to_bytes, to_dict, to_tuple
+from eth_utils import (
+    encode_hex,
+    is_hex,
+    is_integer,
+    is_text,
+    to_bytes,
+    to_dict,
+    to_tuple,
+)
 
 from ddht.abc import RoutingTableAPI, RPCHandlerAPI, RPCRequest
 from ddht.rpc import RPCError, RPCHandler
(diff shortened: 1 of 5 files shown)


ddht/rpc_handlers.py
```diff
@@ -84,9 +87,59 @@ async def do_call(self, params: None) -> NodeInfoResponse:
...
+    params: Any,
+) -> Iterable[Tuple[bytes, Optional[bytes]]]:
+    if not params:
+        raise RPCError("Missing parameters.")
+
+    for kv_pair in params:
+        if len(kv_pair) != 2:
+            raise RPCError(f"Invalid kv_pair length: {len(kv_pair)}.")
+
+        if not is_text(kv_pair[0]):
```
pipermerriam, line 101: I don't think we're going to be able to do everything with unicode encoded bytes since you cannot actually represent the full range of bytes in unicode.  I'm thinking we might want to do:

- key
    - is_text -> must be hexstr
- value
    - is_text -> must be hexstr
    - is_integer -> treat as integer
    - is_none -> `None`

tests/core/test_core_rpc_handlers.py
```diff
@@ -45,6 +53,70 @@ async def test_rpc_nodeInfo(make_request, enr):
...
+async def test_rpc_updateNodeInfo(make_request, enr_manager):
+    # add a kv pair
+    first_response = await make_request("discv5_updateNodeInfo", [("foo", "bar")])
+    assert first_response is None
+    assert enr_manager.enr._kv_pairs[b"foo"] == b"bar"
+    assert enr_manager.enr.sequence_number == 2
+
+    # update a kv pair
+    second_response = await make_request("discv5_updateNodeInfo", [("foo", "xyz")])
+    assert second_response is None
```
pipermerriam, line 66: What do you think about returning the updated ENR from `discv5_updateNodeInfo`

pipermerriam (approved):

tests/core/test_core_rpc_handlers.py
njgheorghita: Yeah that seems reasonable - I was just following the API from `ENRManager.update` here.

pipermerriam: Back when I wrote `ENRManager.update` I certainly considered returning the updated ENR record, but chose against it due to ... the position of Saturn and Mercury being in retrograde?

njgheorghita: I've been there before - it also happens to be the cause of every bug I've written - darn retrograde.

ddht/rpc_handlers.py
```diff
@@ -84,9 +95,62 @@ async def do_call(self, params: None) -> NodeInfoResponse:
...
+        if not kv_pair[1]:
+            value = None
+        elif is_integer(kv_pair[1]):
+            value = to_bytes(kv_pair[1])
+        elif is_text(kv_pair[1]) and is_hex(kv_pair[1]):
+            value = to_bytes(hexstr=kv_pair[1])
+        else:
+            raise RPCError(
+                f"Value: {kv_pair[1]} is type: {type(kv_pair[1])}. "
+                "Values must be text, integer, or None."
```
pipermerriam, line 125:
```suggestion
                "Values must be hex-string, integer, or None."
```

tests/core/test_core_rpc_handlers.py
```diff
@@ -45,6 +54,74 @@ async def test_rpc_nodeInfo(make_request, enr):
...
+    assert second_enr.sequence_number == 3
+
+    # test with multiple kv_pairs, and remove 'foo' key
+    third_response = await make_request(
+        "discv5_updateNodeInfo", [("0xabcd", None), ("0xdef1", 123)]
+    )
+    third_enr = ENR.from_repr(third_response["enr"])
+    assert b"\xab\xcd" not in third_enr._kv_pairs
+    assert third_enr._kv_pairs[b"\xde\xf1"] == to_bytes(123)
+    assert third_enr.sequence_number == 4
```
pipermerriam, line 80: Probably worth having at least one assertion against the `ENRManager.enr` itself to see that the value was actually updated on the other side of the API.  This is mostly just a sanity check.

pipermerriam (approved):

PR #183, opened 2020-11-17
Add support for discv5_lookupENR jsonrpc endpoint

njgheorghita:
## What was wrong?
Add support for `discv5_lookupENR` endpoint.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/rpc_handlers.py
+++ b/ddht/v5_1/rpc_handlers.py
@@ -196,6 +196,28 @@ class DeleteENRHandler(RPCHandler[NodeID, None]):
         return None
 
 
+class LookupENRHandler(RPCHandler[Tuple[NodeID, Optional[int]], GetENRResponse]):
+    def __init__(self, network: NetworkAPI) -> None:
+        self._network = network
+
+    def extract_params(self, request: RPCRequest) -> Tuple[NodeID, Optional[Endpoint], Optional[int]]:
+        raw_params = extract_params(request)
+        if len(raw_params) != 2:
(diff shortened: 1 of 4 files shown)


ddht/v5_1/rpc_handlers.py
```diff
@@ -196,6 +196,27 @@ async def do_call(self, params: NodeID) -> None:
         return None
 
 
+class LookupENRHandler(RPCHandler[Tuple[NodeID, Optional[int]], GetENRResponse]):
+    def __init__(self, network: NetworkAPI) -> None:
+        self._network = network
+
+    def extract_params(self, request: RPCRequest) -> Tuple[NodeID, Optional[int]]:
+        raw_params = extract_params(request)
+        raw_destination = raw_params[0]
```
pipermerriam, line 205: This can raise an `IndexError` since we haven't already validated the number of parameters (input validation is tedious...)

ddht/v5_1/rpc_handlers.py
```diff
@@ -196,6 +196,27 @@ async def do_call(self, params: NodeID) -> None:
...
+        node_id, _ = validate_and_extract_destination(raw_destination)
```
pipermerriam, line 206: Odd to me that we're discarding the `endpoint` part of the return value.  If this is a node we haven't seen yet (and thus it isn't in our ENR database) this will make the lookup go a log slower because it'll cause it to have to fallback to a `recursive_find_nodes` approach.

pipermerriam (approved):

PR #184, opened 2020-11-17
Use 'content' as variable name for proofs

pipermerriam:
## What was wrong?

The code was inconsistently using either `data` or `content` to refer to the raw content of network data.

## How was it fixed?

Unified to use `content` in all places.

### DIFF
--- a/tests/core/v5_1/alexandria/test_partial_proofs.py
+++ b/tests/core/v5_1/alexandria/test_partial_proofs.py
@@ -16,17 +16,17 @@ from ddht.v5_1.alexandria.sedes import ByteList, content_sedes
 
 
 @settings(max_examples=1000)
-@given(data=st.binary(min_size=0, max_size=GB))
-@example(data=b"")
-@example(data=b"\x00" * 31)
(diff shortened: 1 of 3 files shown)


tests/core/v5_1/alexandria/test_partial_proofs.py
```diff
@@ -37,10 +37,10 @@ def test_ssz_full_proofs(data):
     assert len(proven_data_segments) == 1
     start_index, proven_data = proven_data_segments[0]
     assert start_index == 0
-    assert proven_data == data
+    assert proven_data == content
```
marcgarreau, line 40: 👍 for consistency, but are you making a distinction between `content` and `data`? e.g., why `get_proven_data` isn't `get_proven_content`?

marcgarreau (approved):

tests/core/v5_1/alexandria/test_partial_proofs.py
pipermerriam: My mental model has been: `content` is comprised of `data` so trying to only use `content` to mean the full thing.

PR #185, opened 2020-11-17
Include advertisement_radius in Ping and Pong messages

pipermerriam:
depends on #200 #239 #190 #198 

## What was wrong?

Nodes need to be regularly updating others on their advertisement radius.

## How was it fixed?

Include advertisement radius in the Ping and Pong messages.

#### Cute Animal Picture

RaccoonMingoWildlifeRefuge-1


lithp (approved):
I only looked at the last commit, because I think it's the only one not included in a previous PR, but it looks good to me

PR #186, opened 2020-11-17
Implement Advertise/Ack message sending

pipermerriam:
Based on #196 

## What was wrong?

We need to be able to send advertisements around the network.

## How was it fixed?

Implement advertisement message sending APIs on network and client models.

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -30,8 +30,11 @@ from ddht.request_tracker import RequestTracker
 from ddht.subscription_manager import SubscriptionManager
 from ddht.v5_1.abc import NetworkAPI
 from ddht.v5_1.alexandria.abc import AlexandriaClientAPI
+from ddht.v5_1.alexandria.advertisements import Advertisement, partition_advertisements
 from ddht.v5_1.alexandria.constants import ALEXANDRIA_PROTOCOL_ID, MAX_PAYLOAD_SIZE
 from ddht.v5_1.alexandria.messages import (
+    AckMessage,
+    AdvertiseMessage,
     AlexandriaMessage,
     ContentMessage,
     FindNodesMessage,
@@ -43,6 +46,7 @@ from ddht.v5_1.alexandria.messages import (
     decode_message,
 )
 from ddht.v5_1.alexandria.payloads import (
+    AckPayload,
     ContentPayload,
     FindNodesPayload,
     FoundNodesPayload,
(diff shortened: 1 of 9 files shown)


pipermerriam:
only the second commit is relevant.  See #196 for first commit PR

pipermerriam:
This has been rebased and now stands on it's own.  It might have a flaky test failure that is fixed in #206

ddht/v5_1/alexandria/client.py
```diff
@@ -431,6 +461,25 @@ async def get_content(
...
+        advertisements: Collection[Advertisement],
+    ) -> Tuple[AckMessage, ...]:
+        advertisement_batches = partition_advertisements(
+            advertisements, max_payload_size=MAX_PAYLOAD_SIZE,
+        )
+        messages = tuple(AdvertiseMessage(batch) for batch in advertisement_batches)
+        assert len(messages)
+        return tuple(
+            [
+                await self._request(node_id, endpoint, message, AckMessage)
```
marcgarreau, line 478: context gathering: `ACK` is only used to respond to advertisements? obviously a request for content is acknowledged by returning that content, but why is it important to ack the ad?

marcgarreau (approved):

ddht/v5_1/alexandria/client.py
pipermerriam: I'm not sure that the ack is required.  

The original reason that I implement the `Ack` message was to communicate the recipient's updated `advertisement_radius` back to the sender of the advertisement.  The sender uses the `advertisement_radius` to determine whether a given advertisement is *interesting* to a node.  Supposing that I sent you an advertisement that caused something else to get evicted from your advertisement database, your advertisement radius would shrink, and subsequent ads might no longer be interesting to you.  In a perfect would, this is an effective mechanism to prevent someone sending a bunch of advertisements to a node that the node doesn't care about due to a full advertisement db and the ads being less proximate than their current ones.

But... it's looking a lot like validation of an advertisement might take longer than is reasonable to wait for the ack, meaning that the receiving node will need to ack the advertisement before actually validating it, or they sending node will timeout the request/response waiting for the ack due to validation taking too long on the recipient's end.

Since timing out waiting for the Ack isn't ideal, I've got things currently implemented such that the ack is sent prior to full validation...  Which is still better than nothing since it can still give the sending node an updated view of my advertisment radius, but... once the advertisment validates the radius might shrink and the sending node won't know that until they either send another advertisment **or** do a ping/pong.

PR #188, opened 2020-11-17
Add NewType for ContentID

pipermerriam:
Builds on #187 

## What was wrong?

We don't want to mix and match `content_id` values and `node_id` values, nor do we want to overload `Hash32` nor do we want to incidentally pass in a `bytes` type of the wrong type as a `content_id`

## How was it fixed?

Add a `NewType("ContentID")` that we can use for the type hint for `content_id` values.

#### Cute Animal Picture

racoon-bird-feeder-1

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -3,7 +3,7 @@ from typing import Any, AsyncContextManager, Collection, Optional, Sequence, Tup
 
 from async_service import ServiceAPI
 from eth_enr import ENRAPI, ENRDatabaseAPI, ENRManagerAPI
-from eth_typing import Hash32, NodeID
+from eth_typing import NodeID
 import trio
 
 from ddht.abc import RequestTrackerAPI, RoutingTableAPI, SubscriptionManagerAPI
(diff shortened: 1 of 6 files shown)


ddht/v5_1/alexandria/abc.py
```diff
@@ -87,9 +88,9 @@ async def send_get_content(
         node_id: NodeID,
         endpoint: Endpoint,
         *,
-        content_id: Hash32,
+        content_id: ContentID,
         start_chunk_index: int,
-        num_chunks: int,
+        max_chunks: int,
```
marcgarreau, line 93: `num` -> `max` is to account for padding?

marcgarreau (approved):

ddht/v5_1/alexandria/abc.py
pipermerriam: Changed to `max_` because it better matches the semantics of the way the content is provided.  When we request data we say "Give me data starting at chunk index `X` and up to `N` chunks after that".  The only firm requirement is that the responding proof contains the chunk at index `X`.  

So `max_chunks` seemed like a more appropriate name.... but I'm open to alternative suggestions since I recognize that `max_` also suggest that this will be validated which it isn't (and doesn't need to be).

PR #189, opened 2020-11-17
Make `recursive_find_nodes` and `get_nodes_near` use a common protocol based implementation.

pipermerriam:
## What was wrong?

We need the `get_nodes_near` API in Alexandria.  However, previously, copy/paste approach is problematic as these APIs evolve.

## How was it fixed?

Created a `NetworkProtocol` type and extracted the logic for both `recursive_find_nodes` and `get_nodes_near` to use the same logic.

#### Cute Animal Picture

raccoon_feeder

### DIFF
--- a/ddht/v5_1/abc.py
+++ b/ddht/v5_1/abc.py
@@ -1,6 +1,15 @@
 from abc import ABC, abstractmethod
 import logging
-from typing import Any, AsyncContextManager, Collection, Optional, Sequence, Tuple, Type
+from typing import (
+    Any,
+    AsyncContextManager,
+    Collection,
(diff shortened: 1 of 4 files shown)


pipermerriam:
Going to try and deduplicate these.  too much copy-pasta.

pipermerriam:
@lithp using a `typing.Protocol` now which ended up being quite clean (and better than my previous attempt #137 )

pipermerriam:
I'm merging this without review since I'm happy with the protocol approach and it doesn't change functionality.  happy to debate this if someone doesn't like it.

PR #192, opened 2020-11-17
Update Alexandria message API for custom payload encoding

pipermerriam:
## What was wrong?

We need the ability to customize how message payloads are encoded to support the message format of the Advertise message that is being introduced in #186 

## How was it fixed?

These is now a `get_payload_for_encoding()` API on the `AlexandriaMessage` class which allows for custom handling of payload encoding.

#### Cute Animal Picture

istockphoto-598177190-612x612


pipermerriam:
@marcgarreau @lithp merging as this one shouldn't be overly controversial.  See #186 for how this actually gets used...

PR #193, opened 2020-11-17
Alexandria: Add nodes to routing table which ping us

pipermerriam:
## What was wrong?

When a node pings us, they should be a candidate for the routing table

## How was it fixed?

Now when a node pings us, we add them to the routing table.

#### Cute Animal Picture

page-raccoons


njgheorghita (approved):
Lookin good!

PR #196, opened 2020-11-18
Implement Advertisement data model

pipermerriam:
## What was wrong?

We need a data model for advertisements.

## How was it fixed?

Implemented a data model as well as basic utilities for creating and validating.

#### Cute Animal Picture

1d374c942dae6f95a8458c4b93abc6dd

### DIFF
--- a/ddht/v5_1/alexandria/advertisements.py
+++ b/ddht/v5_1/alexandria/advertisements.py
@@ -1,0 +1,195 @@
+import datetime
+import operator
+from typing import Iterable, NamedTuple, Optional, Sequence, Tuple
+
+from eth_keys import keys
+from eth_keys.exceptions import BadSignature
+from eth_typing import Hash32, NodeID
+from eth_utils import keccak, to_tuple
+from eth_utils.toolz import accumulate, cons, sliding_window
+
+from ddht.v5_1.alexandria.content import content_key_to_content_id
+from ddht.v5_1.alexandria.typing import ContentID, ContentKey
+
+ADVERTISEMENT_DOMAIN = b"alexandria-advertisement"
+
+
+def create_advertisement_signing_message(
+    content_id: ContentID, hash_tree_root: Hash32, expires_at: datetime.datetime,
+) -> bytes:
+    return b":".join(
+        (
+            ADVERTISEMENT_DOMAIN,
+            content_id,
+            hash_tree_root,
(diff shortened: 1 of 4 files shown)


pipermerriam:
This should be stable and ready for review.

ddht/v5_1/alexandria/advertisements.py
```diff
@@ -0,0 +1,195 @@
...
+def partition_advertisements(
+    advertisements: Sequence[Advertisement], max_payload_size: int,
+) -> Iterable[Tuple[Advertisement, ...]]:
+    encoded_sizes = tuple(
+        len(advertisement.content_key) + ADVERTISEMENT_FIXED_SIZE
+        for advertisement in advertisements
+    )
+    partition_indices = _get_partition_indices(encoded_sizes, max_payload_size)
+    for left, right in sliding_window(2, partition_indices):
+        yield tuple(advertisements[left:right])
```
lithp, line 195: If I remember correctly you wrote very similar logic in a previous PR. It might simplify things to pull the "break this sequence of items into chunks of at most size n" into a helper method both can call

tests/core/v5_1/alexandria/test_advertisements.py
```diff
@@ -0,0 +1,107 @@
...
+
+    for batch in batches:
+        ssz_payload = tuple(ad.to_sedes_payload() for ad in batch)
+        encoded = ssz.encode(ssz_payload, sedes=AdvertiseSedes)
+        assert len(encoded) <= 512
+
+    for left, right in sliding_window(2, batches):
+        ssz_payload = tuple(ad.to_sedes_payload() for ad in left + (right[0],))
+        encoded = ssz.encode(ssz_payload, sedes=AdvertiseSedes)
+        assert len(encoded) > 512
```
lithp, line 107: Neat, this is property based testing at its best!

lithp (approved):
no complaints here

ddht/v5_1/alexandria/advertisements.py
pipermerriam: Conceptually yes, we do the same thing with ENR records

https://github.com/ethereum/ddht/blob/fd5dd92933d336b49383808c2c78eb7e28159b0c/ddht/enr.py#L10-L46

But the logic is different enough that they don't DRY well.

PR #197, opened 2020-11-18
Implement BroadcastLogAPI

pipermerriam:
Builds on #196 

## What was wrong?

One of the gossip rules for advertisements is that we should only send an advertisement to a peer if we have not "recently" sent it to them.

## How was it fixed?

Implements the `BroadcastLogAPI` and `BroadcastLog` which exposes an API for tracking the last time we sent an advertisement to a given node and for querying this data.

#### Cute Animal Picture

raccoon-herd

### DIFF
--- a/ddht/v5_1/alexandria/broadcast_log.py
+++ b/ddht/v5_1/alexandria/broadcast_log.py
@@ -1,0 +1,37 @@
+import hashlib
+import time
+from typing import Dict
+
+from eth_typing import Hash32, NodeID
+from lru import LRU
+
+from ddht.v5_1.alexandria.advertisements import Advertisement
+from ddht.v5_1.alexandria.constants import ONE_HOUR
+
+
+class BroadcastLog:
+    """
+    Tracks the last time each advertisement was broadcast to each peer.
+    """
+
+    def __init__(self, max_records: int = 8192):
+        self._log: Dict[Hash32, int] = LRU(max_records)
+
+    def log(self, node_id: NodeID, advertisement: Advertisement) -> None:
+        key = Hash32(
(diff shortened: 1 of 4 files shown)


pipermerriam:
This has been rebased and now stands on it's own.  It might have a flaky test failure that is fixed in #206

ddht/v5_1/alexandria/broadcast_log.py
```diff
@@ -0,0 +1,37 @@
...
+
+class BroadcastLog:
+    """
+    Tracks the last time each advertisement was broadcast to each peer.
+    """
+
+    def __init__(self, max_records: int = 8192):
+        self._log: Dict[Hash32, int] = LRU(max_records)
+
+    def log(self, node_id: NodeID, advertisement: Advertisement) -> None:
```
marcgarreau, line 20: curious if recent thoughts on `ACK` affect how you're thinking about these logs, e.g., if no `ACK` was received, do we still log the broadcast/not ping them again?

marcgarreau (approved):

ddht/v5_1/alexandria/broadcast_log.py
pipermerriam: If we didn't receive an `Ack` then the call to `network.advertise(...)` would raise a `trio.TooSlow` and so we wouldn't end up making the call to `BroadcastLog.log(...)` so, no, it would not get broadcast and on some subsequent *loop* over the advertisements we'd re-broadcast, yes.  (this is just how the code is written today)

https://github.com/ethereum/ddht/blob/6bb7cb76afb8f9902f929f309d3feece6e1e1d90/ddht/v5_1/alexandria/advertisement_manager.py#L124-L136

PR #198, opened 2020-11-18
Implement content transmission on AlexandriaNetworkAPI

pipermerriam:
replaces #165 

## What was wrong?

We need high level APIs for retrieving content in Alexandria

## How was it fixed?

Implemented `AlexandriaNetworkAPI.get_content_proof` and `AlexandriaNetworkAPI.get_content_from_nodes`

#### Cute Animal Picture

raccoon-nasty-face

### DIFF
--- a/ddht/v5_1/alexandria/network.py
+++ b/ddht/v5_1/alexandria/network.py
@@ -5,9 +5,11 @@ from typing import Collection, List, Optional, Tuple
 from async_service import Service
 from eth_enr import ENRAPI, ENRManagerAPI, QueryableENRDatabaseAPI
 from eth_enr.exceptions import OldSequenceNumber
-from eth_typing import NodeID
+from eth_typing import Hash32, NodeID
+from eth_utils import ValidationError
 from eth_utils.toolz import cons, first
 from lru import LRU
+from ssz.constants import CHUNK_SIZE
 import trio
 
 from ddht._utils import every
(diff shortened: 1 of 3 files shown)


marcgarreau:
Where does this stand after https://github.com/ethereum/ddht/pull/215?

pipermerriam:
This has been rebased and should be ready for review

tests/core/v5_1/alexandria/test_alexandria_network.py
```diff
@@ -102,3 +105,126 @@ async def test_alexandria_network_responds_to_find_nodes(
...
+
+
+@pytest.mark.trio
+async def test_alexandria_network_get_content_proof_api(
+    alice, bob, alice_alexandria_network, bob_alexandria_client,
+):
+    content = ContentFactory(length=16 * 32 * 4)
+    proof = compute_proof(content, sedes=content_sedes)
+
+    async with bob_alexandria_client.subscribe(GetContentMessage) as subscription:
```
marcgarreau, line 117: `GetContentMessage` handles for `get_content_from_nodes` and `get_content_proof`? What's the thinking behind not having a separate proof message type?

pipermerriam: `get_content_proof` is the mid-level primative for retrieving a proof for some content from a single node.

`get_content_from_nodes` is used to retrieve the full content.  It assumes you have figured out all the nodes that have the content you want.  Internally it uses `get_content_proof` to retrieve the content in pieces, re-assembling it locally, and returning the full content once it has been fully retrieved.

pipermerriam: Oh, as for the separation of not having something like `RawContent` and `ContentProof` message types...  It's just really simple to have a single boolean flag and use the same message type.

We also don't have infrastructure for a request that has two different response types.

And we also can't know ahead of time if the content we're requesting will fit inside a single message or need to be split across multiple messages in a proof...

ddht/v5_1/alexandria/network.py
```diff
@@ -153,6 +161,242 @@ async def find_nodes(
...
+            start_chunk_index=start_chunk_index,
+            max_chunks=max_chunks,
+            request_id=request_id,
+        )
+        if response.payload.is_proof:
+            proof = Proof.deserialize(
+                data=response.payload.payload, sedes=content_sedes,
+            )
+        else:
+            proof = compute_proof(
```
marcgarreau, line 191: fyi this logic fork is not yet triggered in tests or benchmarks

marcgarreau (approved):

ddht/v5_1/alexandria/network.py
pipermerriam: As measured by.... did you run code coverage or something?

pipermerriam: Ah, nevermind, I see it.

marcgarreau: Lower-tech :) while manually stepping through stuff, a log/debug was never executed

pipermerriam:
test failure is unrelated.

PR #199, opened 2020-11-18
Implement NewType for ContentKey

pipermerriam:
## What was wrong?

We have been using `bytes` as the type for content keys.  This is prone to certain types of mistakes that are harder to make if we use a `NewType`

## How was it fixed?

Changed all `content_key` usage to use a new `ContentKey` type which is just a `NewType("ContentKey", bytes)`

#### Cute Animal Picture

8694995818_b22314b967_z

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -18,7 +18,7 @@ from ddht.v5_1.alexandria.messages import (
     TAlexandriaMessage,
 )
 from ddht.v5_1.alexandria.payloads import PongPayload
-from ddht.v5_1.alexandria.typing import ContentID
+from ddht.v5_1.alexandria.typing import ContentKey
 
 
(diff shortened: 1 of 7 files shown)


marcgarreau (approved):
was expecting to see a straight swap from `content_key` to `ContentKey`, but several `id`s are migrated to `key`s. what's the context?

pipermerriam:
I made a mistake in the content retrieval API using `content_id` when we really need to send `content_key` because that is how we're referencing content in all contexts other than determining where advertisements should live.

pipermerriam:
FYI, I'm being a bit more aggressive about merging these than I typically would.  I'm **always** happy to do post merge cleanup.  I'm just trying to optimize for forward progress knowing that there's a lot of this that will need to shift and evolve as it comes together so prioritizing forward progress over (our very high bar for) quality.

marcgarreau:
🤙 same page

PR #191, opened 2020-11-17
Implement ContentStorageAPI

pipermerriam:
Based on #199 

## What was wrong?

We need a way to store content, both in a persistent way, and ephemeral manner.

## How was it fixed?

Implemented the `ContentStorageAPI`.

This includes both an in-memory version and a version that uses the filesystem for persistence.

#### Cute Animal Picture

coonfeeder1-745963

PR #200, opened 2020-11-18
Implement AdvertisementDatabaseAPI

pipermerriam:
This is based on a bunch of other PRs:  #191  #190 #199  #196 #186 

## What was wrong?

We need to store advertisements in a manner that allows for a few different types of queries.

## How was it fixed?

Implemented `AdvertisementDatabaseAPI` which supports persistent storage and retrieval patterns for managing advertisments.

#### Cute Animal Picture

7532425804_c1bbe4dfa3_z

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -32,7 +32,7 @@ from ddht.v5_1.alexandria.messages import (
 )
 from ddht.v5_1.alexandria.partials.proof import Proof
 from ddht.v5_1.alexandria.payloads import AckPayload, PongPayload
-from ddht.v5_1.alexandria.typing import ContentKey
+from ddht.v5_1.alexandria.typing import ContentID, ContentKey
 
 
 class ContentStorageAPI(ABC):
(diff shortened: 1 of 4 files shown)


ddht/v5_1/alexandria/abc.py
```diff
@@ -239,6 +285,8 @@ class AlexandriaNetworkAPI(ServiceAPI, TalkProtocolAPI):
     client: AlexandriaClientAPI
     routing_table: RoutingTableAPI
 
+    advertisement_db: AdvertisementDatabaseAPI
```
marcgarreau, line 288: was expecting to see the `Storage` pattern, like `ContentStorageAPI`, so that an in-memory option could also be used. does that not work here for some reason?

marcgarreau (approved):

ddht/v5_1/alexandria/abc.py
pipermerriam: The "in memory" version is just one that uses `sqlite3.connect(":memory:")` which creates an in-memory sqlite3 database.

PR #201, opened 2020-11-18
Implement AdvertisementManagerAPI

pipermerriam:
## What was wrong?

When we receive an Advertisement it needs to be processed in accordance with the rules set out in the specification to both validate it, **and** to re-broadcast it to the other nodes that may be interested in it.

## How was it fixed?

Implemented `AdvertisementManagerAPI` which listens for incoming advertisements and runs them through the various validation checks prior to re-broadcasting them to peers.

#### Cute Animal Picture

### DIFF
--- a/ddht/v5_1/alexandria/advertisement_manager.py
+++ b/ddht/v5_1/alexandria/advertisement_manager.py
@@ -1,0 +1,257 @@
+import logging
+import secrets
+
+from async_service import Service
+from eth_utils import ValidationError
+import ssz
+from ssz.constants import CHUNK_SIZE
+import trio
+
+from ddht.base_message import InboundMessage
+from ddht.v5_1.alexandria.abc import (
+    AdvertisementDatabaseAPI,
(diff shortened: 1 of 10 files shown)


ddht/v5_1/alexandria/advertisement_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                "Multiple hash tree roots found: content_id=%s  roots=%s",
+                advertisement.content_id.hex(),
+                known_hash_tree_roots,
+            )
+
+        if (
+            known_hash_tree_roots
+            and advertisement.hash_tree_root not in known_hash_tree_roots
+        ):
+            # TODO: perform full validation of the content to determine the
```
pipermerriam, line 96: This TODO will be addressed later as its a situation that should only occur in adversarial situations and until we have the content management built its difficult to write this logic.

pipermerriam:
rage merge

PR #202, opened 2020-11-18
Implement RadiusTracker

pipermerriam:
depends on #185 #239 #200 #198 #190 

## What was wrong?

In order to conform to the network rules for advertisement gossip we need to keep track of the advertisement radius of the other nodes on the network.

## How was it fixed?

Implement `RadiusTrackerAPI` which is a service that runs under the `AlexandriaNetworkAPI` which tracks the three messages which include the advertisement radius (Ping, Pong, Ack) and exposes and API for querying the radius of a node on the network.

#### Cute Animal Picture

file01-11-2017-ykjmecjib5k

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -85,6 +85,16 @@ class ContentProviderAPI(ServiceAPI):
         ...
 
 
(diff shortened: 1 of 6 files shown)


marcgarreau (approved):
lgtm 🚢

PR #206, opened 2020-11-19
Refactor get_nodes_near into explore_target

pipermerriam:
## What was wrong?

The `NetworkAPI.get_nodes_near` was still just kind of terrible.

The `NetworkAPI.recursive_find_nodes` could be better.

## How was it fixed?

It now works very similarly to `recursive_find_nodes` algorithm in that it continually tries to step closer to the `target`, tracking a set of nodes that have already been queried.

The new algorithm differs in that it casts a wider *net* round the target and spends more time exploring that area of the network until either:

- it has queried all of the available nodes
- it has run through multiple rounds without finding improved results.

I also renamed the method `get_nodes_near -> explore_target`.  I don't know how I feel about this but the named seemed marginally better.  I didn't like `near` because the method doesn't necessarily return nodes that are *near*, only the *nearest* it can find.

The `recursive_find_nodes` algorithm was also updated to expand the number of buckets that we query which simply improves the efficiency of the algorithm.

#### Cute Animal Picture

AR-181108128

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -21,6 +21,8 @@ from ddht.kademlia import (
     compute_distance,
     compute_log_distance,
     iter_closest_nodes,
+    sorted_closest,
+    sorted_furthest,
 )
 from ddht.v5_1.abc import (
     ClientAPI,
@@ -40,19 +42,6 @@ from ddht.v5_1.messages import (
     TalkRequestMessage,
 )
 
-NEIGHBORHOOD_DISTANCES = (
-    # First bucket is combined (128 + 64 + 32) since these will rarely be
-    # occupied.
-    tuple(range(1, 224)),
-    # Next few buckets drop in size by about half each time.
-    tuple(range(224, 240)),
-    tuple(range(240, 248)),
-    (248, 249, 250, 251),
-    (252, 253, 254),
-    # This last one is 3/4 of the network
-    (255, 256),
-)
-
 
 async def common_recursive_find_nodes(
     network: NetworkProtocol, target: NodeID
@@ -67,13 +56,15 @@ async def common_recursive_find_nodes(
     async def do_lookup(node_id: NodeID) -> None:
         queried_node_ids.add(node_id)
 
+        distances: Tuple[int, ...]
         if node_id == target:
-            distance = 0
+            distances = (0,)
         else:
-            distance = compute_log_distance(node_id, target)
+            relative_distance = compute_log_distance(node_id, target)
+            distances = tuple(range(1, min(257, relative_distance + 1)))
 
         try:
-            enrs = await network.find_nodes(node_id, distance)
+            enrs = await network.find_nodes(node_id, *distances)
         except trio.TooSlowError:
             unresponsive_node_ids.add(node_id)
             return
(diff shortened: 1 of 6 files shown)


pipermerriam:
Got this crash last night.  Need to figure out how/why

```
  Traceback (most recent call last):
    File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 607, in _probe_network
      nursery.start_soon(self._bond, enr.node_id, endpoint)
    File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/trio/_core/_run.py", line 741, in __aexit__
      raise combined_error_from_nursery
    File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 602, in _probe_network
      found_enrs = await self.explore_target(target_node_id)
    File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 495, in explore_target
      return await common_explore_target(self, target, max_nodes=max_nodes)
    File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 270, in common_explore_target
      last_average_distance = average_distance
    File "/home/piper/python-environments/ddht/lib/python3.8/site-packages/trio/_core/_run.py", line 741, in __aexit__
      raise combined_error_from_nursery
    File "/home/piper/projects/ddht/ddht/v5_1/network.py", line 225, in _worker
      in_flight_node_ids.remove(node_id)
  KeyError: b'\xeb\xa4.\xfc\x93\xa2g\xe5\xc2\xe1\x0f\x08\x89\n\x91\xec)\x1es\xd1\t\xbc]q\xae\xd0c\x0b\xab\xef\xcbr'
```

pipermerriam:
This is stable and ready for review.  It also fixes a flaky test which will be nice to get in sooner than later so I can stop having to kick the CI runs.

ddht/v5_1/network.py
```diff
@@ -130,67 +121,161 @@ async def do_lookup(node_id: NodeID) -> None:
...
+    seen_node_ids.update(node_ids_to_query)
 
-    node_ids_from_routing_table = tuple(network.routing_table.iter_nodes_around(target))
-    node_ids_from_recursive_find = tuple(
-        enr.node_id
-        for enr in enrs_from_recursive_find
-        if enr.node_id != network.local_node_id
-    )
+    # Track what node ids we have already queried
+    queried_node_ids: Set[NodeID] = {network.local_node_id}
```
marcgarreau, line 146: nit: hard to keep track of all these ID sets, `seen`, `to_query`, `queried`, `in_flight` (i think that's all of the states?). i think the distinction between `seen` and `to_query` is least clear. might be worth documenting this lifecycle in the docstring.

ddht/v5_1/network.py
```diff
@@ -404,10 +489,10 @@ async def _fetch_enr(
     async def recursive_find_nodes(self, target: NodeID) -> Tuple[ENRAPI, ...]:
         return await common_recursive_find_nodes(self, target)
 
-    async def get_nodes_near(
+    async def explore_target(
```
marcgarreau, line 492: nit: i understand the reasoning in the PR description, but `explore_target` is pretty mysterious as to what return values to expect. haven't had any ideas i love - most are similar to the name it replaced. maybe just `get/find_nodes` if you think the `near` part is misleading. `near` still feels correct in the relative sense; i don't think there's necessarily an implied promise of `nearest`.

marcgarreau (approved):

pipermerriam:
After doing some experimentation with this today I'm less convinced this API is actually useful.  Right now, the amount of time it takes to complete is completely unreasonable (>2 minutes in the live network).   I'm seeing similarly long call times for `recursive_find_nodes` as well so I'm planning to do some investigative work tomorrow to understand where all the time is being spent.

pipermerriam:
killing this for now https://github.com/ethereum/ddht/pull/215

I did diagnose that some of the problem here was due to cases where we concurrently make 3 requests, two return quickly... one hangs for 10 seconds until it times out.

PR #209, opened 2020-11-19
Add LRU dict to store sessions in a pool

### LINKED ISSUE #84: Change session store to use LRU
## What was wrong?

One aim of the protocol is to safely run on low resource devices with a relatively fixed resource usage.  The current `Pool` implementation for v5.1 uses a timeout mechanism but has no real upper bounds on the number of active sessions.

## How can it be fixed?

Look into using an LRU cache to store sessions.  We can also probably optimize the session objects some using `__slots__` as well as deleting any values that we hold onto (like the WhoAreYou packet) after they are no longer needed.

njgheorghita:
## What was wrong?
Fixes #84 

Updates sessions to only timeout if handshake **has not** been completed. If the handshake has been completed - they cannot timeout. 

## How was it fixed?

Summary of approach.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/pool.py
+++ b/ddht/v5_1/pool.py
@@ -33,6 +33,7 @@ class Pool(PoolAPI):
         enr_db: ENRDatabaseAPI,
         outbound_envelope_send_channel: trio.abc.SendChannel[OutboundEnvelope],
         inbound_message_send_channel: trio.abc.SendChannel[AnyInboundMessage],
+        session_cache_size: int,
         message_type_registry: MessageTypeRegistry = v51_registry,
         events: EventsAPI = None,
     ) -> None:
@@ -46,7 +47,7 @@ class Pool(PoolAPI):
             events = Events()
         self._events = events
 
-        self._sessions = LRU(128, callback=self._evict_session)
+        self._sessions = LRU(session_cache_size, callback=self._evict_session)
         self._sessions_by_endpoint = collections.defaultdict(set)
 
         self._outbound_packet_send_channel = outbound_envelope_send_channel
--- a/ddht/v5_1/pool.py
+++ b/ddht/v5_1/pool.py
@@ -6,6 +6,7 @@ import uuid
 from eth_enr import ENRDatabaseAPI
 from eth_keys import keys
 from eth_typing import NodeID
+from lru import LRU
 import trio
 
 from ddht.base_message import AnyInboundMessage
@@ -45,18 +46,23 @@ class Pool(PoolAPI):
             events = Events()
         self._events = events
 
-        self._sessions = {}
+        self._sessions = LRU(128, callback=self._evict_session)
         self._sessions_by_endpoint = collections.defaultdict(set)
 
         self._outbound_packet_send_channel = outbound_envelope_send_channel
         self._inbound_message_send_channel = inbound_message_send_channel
 
+    def _evict_session(self, key: uuid.UUID, value: SessionAPI) -> None:
+        # callback to remove LRU evicted session from _sessions_by_endpoint
+        self._sessions_by_endpoint[value.remote_endpoint].remove(value)
+
     def remove_session(self, session_id: uuid.UUID) -> SessionAPI:
         try:
-            session = self._sessions.pop(session_id)
+            session = self._sessions[session_id]
         except KeyError:
             raise SessionNotFound(f"No session found with id: {session_id}")
 
+        del self._sessions[session_id]
         self._sessions_by_endpoint[session.remote_endpoint].remove(session)
         return session
 
--- a/tests/core/v5_1/test_dispatcher_session_management.py
+++ b/tests/core/v5_1/test_dispatcher_session_management.py
@@ -25,9 +25,10 @@ async def test_dispatcher_detects_handshake_timeout_as_initiator(
                     await alice_client.send_ping(
                         endpoint=bob.endpoint, node_id=bob.node_id
                     )
+
+                    bob_client.get_manager().cancel()
                     _, envelope = await subscription.receive()
                     assert envelope.packet.is_who_are_you
-                    bob_client.get_manager().cancel()
         # Here we know that the session with bob was created and that bob has terminated
         #
         # Now we wait for the timeout to be triggered
--- a/tests/core/v5_1/test_dispatcher_session_management.py
+++ b/tests/core/v5_1/test_dispatcher_session_management.py
@@ -6,70 +6,24 @@ from ddht.v5_1.constants import SESSION_IDLE_TIMEOUT
 
 
 @pytest.mark.trio
-async def test_dispatcher_detects_handshake_timeout_as_initiator(
-    alice, bob, autojump_clock
+async def test_dispatcher_intiator_and_recipient_detects_handshake_timeout(
+    driver, autojump_clock
 ):
-    try:
-        alice.enr_db.set_enr(bob.enr)
-    except OldSequenceNumber:
-        pass
-    try:
-        bob.enr_db.set_enr(alice.enr)
-    except OldSequenceNumber:
-        pass
+    await driver.initiator.send_ping()
 
-    async with alice.client() as alice_client:
-        async with bob.client() as bob_client:
-            async with alice.events.session_created.subscribe_and_wait():
-                async with bob.events.packet_sent.subscribe() as subscription:
-                    await alice_client.send_ping(
-                        endpoint=bob.endpoint, node_id=bob.node_id
-                    )
+    assert driver.initiator.session.is_during_handshake
+    assert driver.recipient.session.is_before_handshake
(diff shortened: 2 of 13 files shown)


njgheorghita:
> We can also probably optimize the session objects .. deleting any values that we hold onto (like the WhoAreYou packet) after they are no longer needed.

Can you point to where we are holding on `WhoAreYou` packets after they're no longer needed? It looks to me like we aren't really storing them anywhere after they're discarded / deemed unnecessary.

ddht/v5_1/pool.py
```diff
@@ -45,18 +46,23 @@ def __init__(
             events = Events()
         self._events = events
 
-        self._sessions = {}
+        self._sessions = LRU(128, callback=self._evict_session)
```
pipermerriam, line 49: I think this should be a configurable parameter that can be controlled from the CLI.  Something like `--session-cache-size`

ddht/v5_1/pool.py
```diff
@@ -45,18 +46,23 @@ def __init__(
...
+        self._sessions_by_endpoint[value.remote_endpoint].remove(value)
+
     def remove_session(self, session_id: uuid.UUID) -> SessionAPI:
         try:
-            session = self._sessions.pop(session_id)
+            session = self._sessions[session_id]
         except KeyError:
             raise SessionNotFound(f"No session found with id: {session_id}")
 
+        del self._sessions[session_id]
```
pipermerriam, line 65: I think you can combine these into:

```python
try:
    del self._sessions[session_id]
except KeyError:
    raise SessionNotFound(...)
```

pipermerriam (commented):
Sorry to not have outlined this better in the issue.  We also want to modify the timeout mechanism for sessions.

Currently, sessions can timeout in two ways.

1. Takes too long for handshake to complete
2. It has been too long since the last message was received at.

I want to get rid of the later.  Once a session has completed the handshake, it is valid indefinitely.  This should allow us to remove the `_last_message_received_at` value.

pipermerriam:
> Can you point to where we are holding on WhoAreYou packets after they're no longer needed? 

Looks like they aren't the `WhoAreYou` packets but here are the two places where a packet/message is stored.  I *think* these can be deleted once the handshake has completed.

- https://github.com/ethereum/ddht/blob/fd5dd92933d336b49383808c2c78eb7e28159b0c/ddht/v5_1/session.py#L274
- https://github.com/ethereum/ddht/blob/fd5dd92933d336b49383808c2c78eb7e28159b0c/ddht/v5_1/session.py#L344-L350

pipermerriam:
There's also more than can be deleted once the session handshake is complete:

These buffers are not needed after they have been processed:

- buffers: https://github.com/ethereum/ddht/blob/fd5dd92933d336b49383808c2c78eb7e28159b0c/ddht/v5_1/session.py#L472-L475
- processed here: https://github.com/ethereum/ddht/blob/fd5dd92933d336b49383808c2c78eb7e28159b0c/ddht/v5_1/session.py#L437-L445

pipermerriam (commented):
I'm going to advocate for you focusing on the timeout issue in this PR and then we can focus on reducing the size of sessions in a subsequent set of PRs.

ddht/v5_1/pool.py
njgheorghita: I'm not sure, now with `self._sessions: LRUDict` - removing the element via `del` does not return the deleted element like it was previously with `pop()`. So we need two steps here - 1 to grab the `session` - 2 to delete the `session` from the `self._sessions: LRUDict`. Then we continue on the next line to remove the `session` from the `self.sessions_by_endpoint` store.

njgheorghita:
@pipermerriam It seems like there's something I'm missing that's causing these hanging tests. Digging a little into this test it seems as though `bob` is still completing the handshake with `alice` causing this and its sibling test to hang indefinitely with the new timeout rules.
https://github.com/ethereum/ddht/blob/ed43f8444258d83973d1d182b93378a347b65d71/tests/core/v5_1/test_dispatcher_session_management.py#L7-L30

pipermerriam:
Try disabling this test:

```
tests/core/v5_1/test_network.py::test_network_get_nodes_near 
```

ddht/v5_1/pool.py
pipermerriam: right...  carry on.  Also, WTF is up with not supporting `pop()` in `LRU`?

tests/core/v5_1/test_pool.py
```diff
@@ -9,6 +9,8 @@
 from ddht.v5_1.pool import Pool
 from ddht.v5_1.session import SessionInitiator, SessionRecipient
 
+TEST_CACHE_SESSION_SIZE = 128
```
pipermerriam, line 12: Might as well drop this even lower me thinks.  No need to perform a full 128 handshakes (384 packets over 256 sockets).  Maybe drop this to something like 16

tests/core/v5_1/test_pool.py
```diff
@@ -38,12 +40,32 @@ async def pool(tester, initiator, events, channels):
...
     )
     tester.register_pool(pool, channels)
     return pool
 
 
+@pytest.mark.trio
+async def test_pool_lru_caches_sessions(tester, events, pool):
+    recipients = []
+    for _ in range(0, 150):
+        recipients.append(tester.node())
```
pipermerriam, line 54: nip

```python
recipients = tuple(tester.node() for _ in range(150))
```

pipermerriam (approved):
test cleanup around session timeouts looks nice.

njgheorghita:
@pipermerriam I've seen that `test_network_get_nodes_near` test act a little flaky before, but it doesn't seem to be related at all to the failure here. 

What seems to be the problem is that `bob_client` survives when we try to kill him (with `bob_client.get_manager().cancel()` and ends up completing the handshake. 
https://github.com/ethereum/ddht/blob/master/tests/core/v5_1/test_dispatcher_session_management.py#L23

Then we end up listening for an `alice.events.session_timeout` event that never comes, b/c `bob_client` has now completed the handshake and is no longer considered an "evictable" session. 

Do you know of a more effective way to kill bob? I've tried using `bob_client.get_manager().stop()` with no luck. I'm 99% this is the problem, because is we insert `await bob_client.get_manager().wait_finished()` we also hang indefinitely - indicating that there is some problem in killing bob. What's the best way to kill bob?

pipermerriam:
You may need to go lower level to avoid the race condition and manually run the Bob side of things.

Sent from ProtonMail mobile

-------- Original Message --------

pipermerriam:
For more context.  What I *think* is happening here is:

The previous test worked because of lucky coroutine shenanigans.  There was enough coroutine switching happening that bob didn't get a chance to finish the handshake before the cancellation took place.  The changes you made unfortunately may have shuffled that around such that the cancel no longer happens fast enough and bob gets a chance to finish the handshake before the cancellation takes place.

I enourage you to explore yourself whether there's a clean way to do this without getting low level but... you very well may need to ditch using `bob_client/bob_network` and instantiate a `SessionXXX` object yourself and feed the messages manually.  There is *some* machinery that might help you.  See the tests here: https://github.com/ethereum/ddht/blob/master/tests/core/v5_1/test_session.py

tests/core/v5_1/test_dispatcher_session_management.py
```diff
@@ -5,58 +5,25 @@
...
-        with trio.fail_after(SESSION_IDLE_TIMEOUT + 1):
-            async with alice.events.session_timeout.subscribe() as subscription:
-                session = await subscription.receive()
-                assert session.remote_node_id == bob.node_id
-
-
-@pytest.mark.trio
-async def test_dispatcher_detects_handshake_timeout_as_recipient(
-    alice, bob, autojump_clock
+async def test_dispatcher_intiator_and_recipient_detects_handshake_timeout(
```
njgheorghita, line 8: @pipermerriam If you don't mind giving one final :eyes: to this test change and double check that I'm not losing any test coverage with this change.

pipermerriam (approved):
We are definitely losing coverage.  The thing we aren't testing anymore is whether we actively evict sessions that timeout during handshakes.  I don't think the loss of coverage is critically important but I do think this is an area that we need to circle back to.  Can you open an issue to to track this coverage loss.

The other thing that I think we're going to run into is that someone can probably force us to flush our sessions by initiating lots and lots of handshakes in quick succession.  It won't crash the app but it will require us to re-establish sessions which costs something.  We can probably mitigate this (in some other PR) by simply limiting the number of incomplete handshakes for a given `Endpoint` so that executing such an attack requires many IP addresses which makes the attack more expensive.

pipermerriam:
@njgheorghita what's the status of this PR?

njgheorghita:
@pipermerriam I've been tinkering locally on trying to improve the test coverage - hoped that maybe inspiration would strike over the weekend, but if not, there's an issue up and I'll push this through on Monday

PR #217, opened 2020-11-20
Bugfix: handle EmptyFindNodesResponse in pong handles

pipermerriam:
## What was wrong?

When a pong happens we lookup their ENR record.  This lookup can result in a network request which can hang for as much as 10 seconds.  The lookup can raise an exception if they fail to return their ENR record.

## How was it fixed?

The processing is now done in a nursery so it doesn't block handling of other incoming messages.

We now also handle the `EmptyFoundNodesResponse` message.

#### Cute Animal Picture

racoon-trash

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -13,6 +13,7 @@ from lru import LRU
 import trio
 
 from ddht._utils import every, reduce_enrs
+from ddht.base_message import InboundMessage
 from ddht.constants import ROUTING_TABLE_BUCKET_SIZE
 from ddht.endpoint import Endpoint
 from ddht.exceptions import DuplicateProtocol, EmptyFindNodesResponse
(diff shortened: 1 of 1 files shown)


ddht/v5_1/network.py
```diff
@@ -449,25 +450,35 @@ async def _manage_routing_table(self) -> None:
...
+        ) -> None:
+            try:
                 enr = await self.lookup_enr(
                     request.sender_node_id,
                     enr_seq=request.message.enr_seq,
                     endpoint=request.sender_endpoint,
                 )
-                self.routing_table.update(enr.node_id)
-                self._routing_table_ready.set()
+            except EmptyFindNodesResponse:
```
marcgarreau, line 462: most failed lookups here are timeouts due to being blocked? any other identified reasons?

marcgarreau (approved):

ddht/v5_1/network.py
pipermerriam: `lookup_enr` is a simple wrapper around doing `FindNodes(0)`.  The `0` distance here has special meaning in the protocol meaning, give me your ENR record.  Mis-behaving/broken/malicious/whatever nodes can return an empty response and this is the exception that will be raised in that case since in all reasonable situations a node can always return their own ENR record.

PR #218, opened 2020-11-20
Harden recursive_find_nodes against timeouts

pipermerriam:
## What was wrong?

The way that `recursive_find_nodes` was implemented was subject to significant slowdowns by unresponsive nodes.

Additionally, many use cases for `recursive_find_nodes` don't actually **require** we complete the lookup.  For example: the usage in `NetworkAPI.lookup_enr` can quit as soon as the target node has been found.  Same with future usage in the alexandria network for things like finding nodes that have the content you want.

### Speedups

First, the lookups are now done with two layers of concurrency.  This adds some resistance so that when a single node is unresponsive, other lookups will still continue to happen in other workers.

Second, the new `adaptive_timeout` helper cancels requests that are taking too long when other requests are returning quickly.  This helps us timeout request sooner than the 10 second boundary, speeding up our happy path as long as there are other nodes that are not timing out.

Last, we keep a cache of unresponsive nodes.  This allows us speed up successive calls to `recursive_find_nodes`, not retrying nodes that recently timed out.

Prior to this, I was seeing RFN take as much as 70 seconds to return.  Now the initial request can still take as much as 20-50 seconds, but once the cache has been populated, requests that don't hit any timeouts now return in 1-3 seconds.

### API changes

The `NetworkAPI.recursive_find_nodes` now returns an async iterator (`trio.abc.ReceiveChannel`).  This allows for aborting the lookup partway through if the caller so desired by simply exiting the async iteration.

#### Cute Animal Picture

12580436_web1_67898147_2495135923858928_7624345396042530816_n

### DIFF
--- a/ddht/_utils.py
+++ b/ddht/_utils.py
@@ -17,6 +17,7 @@ from typing import (
     Collection,
     Iterable,
     Iterator,
+    List,
     Optional,
     Sequence,
     Tuple,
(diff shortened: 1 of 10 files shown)


pipermerriam:
This doesn't need to be blocked by https://github.com/ethereum/ddht/issues/220 but the two might be *somehow* related.  I'm still confused.

pipermerriam:
I merged this without getting review.  I'm happy to do post review cleanup.  It got bigger than I wanted but the fixes in it were significant enough that i wanted to get them in and start building on them because basically all of my other branches ended up needing this stuff.  Primary things are:

- `recursive_find_nodes` now returns an async iterator so that call sites don't have to wait until the full lookup is done to start doing things with the found nodes.  THis gives a *massive* speedup to `NetworkAPI.lookup_enr` because it almost always exits early now.  Similarly, now, if the node information we need happens to be in our routing table, it can completely short circuit all network lookups.
- The actual implementation of RFN is way faster.  Previously I was seeing 60-70s times in live network conditions.  Now I see plenty of 1-2 second calls, and normally only 10-20 seconds when it encounters a timeout.  The previous point helps alleviate the impact of these timeouts on the application.

PR #224, opened 2020-11-23
Jsonrpc talk endpoint support

njgheorghita:
## What was wrong?
Add support for the following jsonrpc endpoints
- `discv5_talk`
- `discv5_sendTalkRequest`
- `discv5_sendTalkResponse`

## How was it fixed?
Added support for the endpoints. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/tools/web3.py
+++ b/ddht/tools/web3.py
@@ -165,6 +165,8 @@ class RPC:
     findNodes = RPCEndpoint("discv5_findNodes")
     sendFindNodes = RPCEndpoint("discv5_sendFindNodes")
     sendFoundNodes = RPCEndpoint("discv5_sendFoundNodes")
+    sendTalkRequest = RPCEndpoint("discv5_sendTalkRequest")
+    sendTalkResponse = RPCEndpoint("discv5_sendTalkResponse")
 
 
 NodeIDIdentifier = Union[ENRAPI, str, bytes, NodeID, HexStr]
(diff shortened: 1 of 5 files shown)


ddht/tools/web3.py
```diff
@@ -343,13 +362,36 @@ class DiscoveryV5Module(ModuleV2):  # type: ignore
         result_formatters=lambda method: find_nodes_response_formatter,
         mungers=[find_nodes_munger],
     )
-    send_find_nodes = Method(
+    send_find_nodes: Method[
```
njgheorghita, line 365: Any idea why mypy was missing these types? They still aren't being correctly validated btw...

njgheorghita:
@pipermerriam Most of the methods that the jsonrpc endpoints point towards accept an optional `endpoint` or `request_id`. Currently, when using jsonrpc, these args default to `None`. Is it worthwhile enhancing the jsonrpc endpoints to accept these optional parameters?

ddht/tools/web3.py
pipermerriam: If you're saying that these type hints aren't actually being validated (i.e. that you could changed one of them arbitrarily and mypy wouldn't complain) then I think that's something we should investigate (probably separately from this PR).  I'd recommend starting by using some `reveal_type` statements to figure out what mypy thinks these types are.

My guess is that web3.py might be leaking `Any` which is typically where this type of deficiency shows up.

pipermerriam (approved):

pipermerriam:
> @pipermerriam Most of the methods that the jsonrpc endpoints point towards accept an optional `endpoint` or `request_id`. Currently, when using jsonrpc, these args default to `None`. Is it worthwhile enhancing the jsonrpc endpoints to accept these optional parameters?

Not crutial.  We can add it later.

PR #235, opened 2020-11-25
Implement messages for Locate/Locations

pipermerriam:
## What was wrong?

We need message pairs for querying the advertisement tables managed by other nodes so that nodes can find the content they are looking for.

## How was it fixed?

Implemented messages for `Locate` and `Locations` messages.

#### Cute Animal Picture

Red-Fox-Jumping

### DIFF
--- a/ddht/v5_1/alexandria/sedes.py
+++ b/ddht/v5_1/alexandria/sedes.py
@@ -26,6 +26,7 @@ class ByteList(List):  # type: ignore
 
 byte_list = ByteList(max_length=2048)
 uint40 = UInt(40)
+content_key_sedes = ByteList(max_length=256)
 
 
 PingSedes = Container(field_sedes=(uint32,))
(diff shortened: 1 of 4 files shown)


ddht/v5_1/alexandria/sedes.py
```diff
@@ -43,5 +44,8 @@ def deserialize(self, value: bytes) -> bytes:
 AdvertiseSedes = List(AdvertisementSedes, max_length=32)
 AckSedes = Container(field_sedes=(uint256,))
 
+LocateSedes = Container(field_sedes=(content_key_sedes,))
+LocationsSedes = Container(field_sedes=(uint8, List(AdvertisementSedes, max_length=32)))
```
njgheorghita, line 48: Is `32` as a max number of advertisements to return is a limit that's up to the implementation to decide?

njgheorghita (approved):
Looks good to me!

ddht/v5_1/alexandria/sedes.py
pipermerriam: It's derived from the maximum number of advertisements that *could* possibly fit into a packet.  Realistically, the max is probably about half of that number.

PR #236, opened 2020-11-25
Implement locate locations message sending

pipermerriam:
builds on #235

## What was wrong?

#235 implements the message types for requesting and sending the locations of content.  We still need message primatives on the `AlexandriaClientAPI` for sending them.

## How was it fixed?

Implemented the standard message sending APIs for both individual message sending and for doing the round trip request/response.

#### Cute Animal Picture

A-dreamstime_xxl_32311527_rjm4k4

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -40,6 +40,8 @@ from ddht.v5_1.alexandria.messages import (
     FindNodesMessage,
     FoundNodesMessage,
     GetContentMessage,
+    LocateMessage,
+    LocationsMessage,
     PingMessage,
     PongMessage,
     TAlexandriaMessage,
@@ -51,6 +53,8 @@ from ddht.v5_1.alexandria.payloads import (
     FindNodesPayload,
     FoundNodesPayload,
     GetContentPayload,
+    LocatePayload,
+    LocationsPayload,
     PingPayload,
     PongPayload,
 )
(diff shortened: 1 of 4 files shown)


ddht/v5_1/alexandria/client.py
```diff
@@ -393,6 +397,37 @@ async def send_ack(
...
+        *,
+        advertisements: Sequence[Advertisement],
+        request_id: bytes,
+    ) -> int:
+        advertisement_batches = partition_advertisements(
+            advertisements, max_payload_size=MAX_PAYLOAD_SIZE,
+        )
+        num_batches = len(advertisement_batches)
+        for batch in advertisement_batches:
+            message = LocationsMessage(LocationsPayload(num_batches, batch))
```
njgheorghita, line 426: https://github.com/ethereum/ddht/blob/3719592a22f2d47754c34b6743f0c0de28ae1b16/ddht/v5_1/alexandria/payloads.py#L62-L64

https://github.com/ethereum/ddht/blob/d1b18da2f96dfb9eb6f849a599ff2cb53c195b21/tests/core/v5_1/alexandria/test_message_encoding.py#L112-L113

Not sure if this is worth changing, but `total` referring to the total # of batches rather than total number of advertisements was somewhat unclear to me.

ddht/v5_1/alexandria/client.py
```diff
@@ -496,6 +531,39 @@ async def advertise(
...
+        ) as subscription:
+            head_response = await subscription.receive()
+            total = head_response.message.payload.total
+            responses: Tuple[InboundMessage[LocationsMessage], ...]
+            if total == 1:
+                responses = (head_response,)
+            elif total > 1:
+                tail_responses: List[InboundMessage[LocationsMessage]] = []
+                for _ in range(total - 1):
+                    tail_responses.append(await subscription.receive())
```
njgheorghita, line 557: What happens if for whatever reason one of the messages gets lost in transmission?

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -469,15 +455,122 @@ async def _respond():
...
+                bob.node_id,
+                bob.endpoint,
+                content_key=advertisements[0].content_key,
+                request_id=b"\x01",
+            )
+            assert len(responses) == 1
+            response = responses[0]
+            assert isinstance(response.message, LocationsMessage)
+            assert response.message.payload.total == 1
+            assert response.message.payload.locations == advertisements
```
njgheorghita, line 529: nitpick, but it looks like the returned `request_id` isn't being tested in this and the following test.

njgheorghita (approved):
Looks good!

ddht/v5_1/alexandria/client.py
pipermerriam: Then the whole request errors out.  It's not ideal because *probably* the incomplete set of other responses contain useful information...  Which suggests that we probably need to do roughly the same pattern here as is done with `NetworkAPI.recursive_find_nodes` and to stream the response ENR records asynchronously.

```python
async with network.find_nodes(...) as enrs_aiter:
    async for enr in enrs_aiter:
        ...
```

But this is subject to all of the necessary exception handling so it turns into:

```python
try:
    async with network.find_nodes(...) as enrs_aiter:
        async for enr in enrs_aiter:
            ...
except (trio.TooSlowError, ValidationError):
    ...
```

I think this would yield a performance improvement to RFN as well since it would allow workers to consume the ENRS as soon as they are received rather than having to wait until all of them have been received.

ddht/v5_1/alexandria/client.py
pipermerriam: I'm sticking with the `total` for now, as it comes from the core discv5 spec terminology used in the base protocol and consumers of these APIs should never actually be dealing with the fields in these messages.

PR #237, opened 2020-11-25
Implement AlexandriaNetworkAPI.locate

pipermerriam:
Builds on #236 

## What was wrong?

#235 and #236 implement the primatives for do location queries down at the `AlexandriaClientAPI`.  We need the high level APIs on `AlexandriaNetworkAPI`.

## How was it fixed?

Implemented `AlexandriaNetworkAPI.locate`

#### Cute Animal Picture

jumping

### DIFF
--- a/ddht/v5_1/alexandria/abc.py
+++ b/ddht/v5_1/alexandria/abc.py
@@ -446,6 +446,17 @@ class AlexandriaNetworkAPI(ServiceAPI, TalkProtocolAPI):
         ...
 
     @abstractmethod
+    async def locate(
+        self,
+        node_id: NodeID,
+        *,
+        content_key: ContentKey,
+        endpoint: Optional[Endpoint] = None,
+        request_id: Optional[bytes] = None,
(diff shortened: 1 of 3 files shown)


ddht/v5_1/alexandria/abc.py
```diff
@@ -445,6 +445,17 @@ async def advertise(
     ) -> Tuple[AckPayload, ...]:
         ...
 
+    @abstractmethod
+    async def locate(
```
marcgarreau, line 449: thought i had a decent picture of it, but now I'm less sure: what is the distinction between the `AlexandriaNetworkAPI` and `AlexandriaClientAPI`?

pipermerriam: Note that the `AlexandriaClientAPI` returns the full response `InboundMessage[LocationsMessage]`.

The `AlexandriaNetworkAPI` unpacks these (and strings them together) to return the actual `Advertisement` instances that are embedded in all of the response messages.

marcgarreau (approved):

PR #245, opened 2020-11-30
Use cache for NetworkAPI.bond

pipermerriam:
## What was wrong?

We should avoid re-bonding with nodes too often..

## How was it fixed?

Add a cache to `NetworkAPI.bond` which will cache bonding for 60 seconds.

#### Cute Animal Picture

3578048015_ccf8e25b89_o


pipermerriam:
I think this is the wrong mechanism.  We're going to instead want to integrate this into the `PoolAPI` but right now it doesn't support the functionality we need (querying by node-id, only grabbing sessions with a completed handshake).

PR #247, opened 2020-12-01
[WIP] Refactor session timeout monitoring

njgheorghita:
## What was wrong?
A more efficient way to timeout sessions was needed, to avoid hitting the max # of child task limit.

## How was it fixed?
Stop monitoring for session timeouts as soon as the handshake is complete.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/ddht/v5_1/session.py
+++ b/ddht/v5_1/session.py
@@ -64,6 +64,7 @@ class BaseSession(SessionAPI):
         self.id = uuid.uuid4()
 
         self.created_at = trio.current_time()
+        self._handshake_complete = trio.Event()
 
         if events is None:
             events = Events()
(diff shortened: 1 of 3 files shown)


ddht/v5_1/session.py
```diff
@@ -159,6 +159,15 @@ def keys(self) -> SessionKeys:
     async def _process_message_buffers(self) -> None:
         ...
 
+    async def await_handshake_completion(self) -> None:
+        if self.is_after_handshake:
+            raise Exception("Handshake is already complete")
+
+        async with self._events.session_handshake_complete.subscribe() as subscription:
```
pipermerriam, line 166: Rather than this, you should use a `trio.Event()`.  This method would simply be:

```python
def __init__(self, ...):
    self._handshake_complete = trio.Event()
async def await_handshake_completion(self) -> None:
    await self._handshake_complete.wait()
```

Then, in the two places where we do `self._status = SessionStatus.AFTER` we can also do `self._handshake_complete.set()`.

pipermerriam (approved):
Assuming tests pass this looks :+1: in its current iteration.

PR #250, opened 2020-12-01
More long running tests

pipermerriam:
## What was wrong?

The generation of random advertisements was taking a long time

## How was it fixed?

Changed the strategy for generating the content keys to be faster.

#### Cute Animal Picture

dog-jumping


pipermerriam:
this dropped a 3s test to 0.3 seconds :smile:

PR #252, opened 2020-12-02
Proper exception handling and testing for found nodes

pipermerriam:
## What was wrong?

An unhandled and untested exception case in the handling of `ClientAPI.find_nodes` when we receive a message with a `total == 0` which is invalid.

## How was it fixed?

Raise a `ValidationError` and add testing.

### DIFF
--- a/ddht/v5_1/client.py
+++ b/ddht/v5_1/client.py
@@ -6,6 +6,7 @@ from async_service import Service
 from eth_enr import ENRAPI, ENRManager, QueryableENRDatabaseAPI
 from eth_keys import keys
 from eth_typing import NodeID
+from eth_utils import ValidationError
 import trio
 
 from ddht.base_message import AnyInboundMessage, AnyOutboundMessage, InboundMessage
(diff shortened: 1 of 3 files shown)


ddht/v5_1/client.py
```diff
@@ -244,6 +245,7 @@ async def send_found_nodes(
             enrs, max_payload_size=FOUND_NODES_MAX_PAYLOAD_SIZE
         )
         num_batches = len(enr_batches)
+        assert enr_batches
```
njgheorghita, line 248: [antipattern](https://github.com/ethereum/snake-charmers-tactical-manual/blob/4dc11553939675e427e5daec51665033b0777538/antipatterns.md#assertions-in-runtime-code)? Is there a special case to use a bare assert here?

njgheorghita (approved):
Looks good, thanks for handling this one - it slipped down my todo list

ddht/v5_1/client.py
pipermerriam: debugging leftover, thanks for catching it.

PR #254, opened 2020-12-03
Detect session timeout event emitted in dispatcher tests

### LINKED ISSUE #228: Improve session timeout testing and add protection for spam handshakes
## What was wrong?
In #209 - sessions will only timeout if the handshake has not been completed. The testing around session timeout is insufficient to determine that we actively timeout a session before the handshake is able to complete. [discussion](https://github.com/ethereum/ddht/pull/209#pullrequestreview-536602797). 

## How can it be fixed?
Update the [test](https://github.com/ethereum/ddht/blob/master/tests/core/v5_1/test_dispatcher_session_management.py#L7) to validate that the proper `session_timeout` event is emitted when a session times out due to a lack of response - for both the recipient and initiator.

This will likely need to be done by manually initiating the handshake and then halting it before completion since the `Client` APIs do not provide a reliable mechanism for halting a handshake midway.

njgheorghita:
## What was wrong?
Fixes #228

In the dispatcher tests, we are only testing that the status of a node is updated correctly in the case of a timeout. To be more robust, the tests should prove that a `session_timeout` event is emitted upon timeout.

## How was it fixed?
Updated the tests to detect session timeouts

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/v5_1/test_dispatcher_session_management.py
+++ b/tests/core/v5_1/test_dispatcher_session_management.py
@@ -1,29 +1,53 @@
+import secrets
+
 from eth_enr import OldSequenceNumber
 import pytest
 import trio
 
+from ddht.base_message import AnyOutboundMessage
 from ddht.v5_1.constants import SESSION_IDLE_TIMEOUT
+from ddht.v5_1.messages import PingMessage
 
 
 @pytest.mark.trio
-async def test_dispatcher_intiator_and_recipient_detects_handshake_timeout(
-    driver, autojump_clock
+async def test_dispatcher_detects_handshake_timeout_as_initiator(
+    alice, bob, alice_client, bob_client, autojump_clock
 ):
-    await driver.initiator.send_ping()
+    await bob_client.get_manager().stop()
+    alice_client.pool.initiate_session(bob.endpoint, bob.node_id)
 
-    assert driver.initiator.session.is_during_handshake
-    assert driver.recipient.session.is_before_handshake
-    assert driver.initiator.session.remote_node_id == driver.recipient.node.node_id
-    assert not driver.initiator.session.is_timed_out
-    assert not driver.recipient.session.is_timed_out
+    with trio.fail_after(SESSION_IDLE_TIMEOUT + 1):
+        async with alice.events.session_timeout.subscribe() as subscription:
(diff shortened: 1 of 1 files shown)


tests/core/v5_1/test_dispatcher_session_management.py
```diff
@@ -1,29 +1,60 @@
...
 
 
 @pytest.mark.trio
-async def test_dispatcher_intiator_and_recipient_detects_handshake_timeout(
-    driver, autojump_clock
+async def test_dispatcher_detects_handshake_timeout_as_initiator(
+    alice, bob, alice_client, bob_client, autojump_clock
 ):
-    await driver.initiator.send_ping()
+    await bob_client.manager.stop()
```
pipermerriam, line 16: The *correct* way to do this externally on a service is `bob_client.get_manager().stop()` but this is almost exclusively only for type checking reasons.

Otherwise, maybe instead of stopping bob, you could just not include the `bob_client` fixture in this test which I think would have the same effect.

tests/core/v5_1/test_dispatcher_session_management.py
```diff
@@ -1,29 +1,60 @@
...
+    message = AnyOutboundMessage(
+        PingMessage(
+            secrets.token_bytes(4), bob_client.enr_manager.enr.sequence_number,
+        ),
+        bob.endpoint,
+        bob.node_id,
+    )
+    await alice_client.dispatcher._outbound_message_send_channel.send(message)
```
pipermerriam, line 24: I assume this line is to trigger creation of the new session.  If so, I think it would be better to directly call `alice_client.pool.initiate_session` (or `alice_client.pool.receive_session`).

tests/core/v5_1/test_dispatcher_session_management.py
njgheorghita: For some reason, the test becomes a little flaky when leaving out `bob_client` and explicitly stopping `bob_client` solves the flake

pipermerriam: In that case, you should instead create a new `enr` record and simply try to connect to it (since it won't have a running service listening on whatever port it generates).

PR #258, opened 2020-12-04
Implement AdvertisementProviderAPI

pipermerriam:
## What was wrong?

We need a process that listens for LOCATE messages and responds with information from our advertisement table

## How was it fixed?

A new `AdvertisementProviderAPI` service which subscriibes to `LocateMessage` and responds with the relevant information from the advertisement database.

#### Cute Animal Picture

show-jumping-rabbit


pipermerriam:
rage merge!

PR #262, opened 2020-12-07
Try to firm up flaky RPC tests

pipermerriam:
## What was wrong?

There was some fixture cleanup in the RPC tests and a subtle mis-use of the `trio.Condition` object used in RFN

## How was it fixed?

Removed the redundant fixture code

Moved the acqusition of the condition lock to the outermost layer to avoid race conditions where the condition is triggered between the time when the lock is released and acquired.

#### Cute Animal Picture

adelie-penguin-jumping-ocean ngsversion 1396530997321 adapt 1900 1

### DIFF
--- a/ddht/v5_1/network.py
+++ b/ddht/v5_1/network.py
@@ -230,11 +230,11 @@ async def common_recursive_find_nodes(
 
     async def _monitor_done(send_channel: trio.abc.SendChannel[ENRAPI]) -> None:
         async with send_channel:
-            while True:
-                # this `fail_after` is a failsafe to prevent deadlock situations
-                # which are possible with `Condition` objects.
-                with trio.fail_after(60):
-                    async with condition:
+            async with condition:
+                while True:
+                    # this `fail_after` is a failsafe to prevent deadlock situations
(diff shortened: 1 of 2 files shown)


ddht/v5_1/network.py
```diff
@@ -230,11 +230,11 @@ async def worker(
...
-            while True:
-                # this `fail_after` is a failsafe to prevent deadlock situations
-                # which are possible with `Condition` objects.
-                with trio.fail_after(60):
-                    async with condition:
+            async with condition:
+                while True:
+                    # this `fail_after` is a failsafe to prevent deadlock situations
+                    # which are possible with `Condition` objects.
+                    with trio.fail_after(60):
```
pipermerriam, line 237: I've been just kicking the CI tests over and over to re-run them to try and find the failures.  So far this PR hasn't had a single timeout.  Not conclusive at all but... I think this is where the flakyness was coming from.

TLDR `Condition` objects seem to be the right tool for certain concurrency situations but they are also easy to use incorrectly (and potentially end up with deadlocks).

PR #263, opened 2020-12-07
Implement stream find nodes api.

### LINKED ISSUE #255: Convert `ClientAPI.find_nodes` and `NetworkAPI.find_nodes` to use async iterators
## What was wrong?

inspired from: https://github.com/ethereum/ddht/pull/236#discussion_r535285375

The `ClientAPI.find_nodes` and `NetworkAPI.find_nodes` involve a single request followed by a multipart response.  The current API implements an all-or-nothing pattern where we discard and error out in cases like receiving 2/3 packets where the 3rd packet never arrives.  In this case, our code would be sitting with the information from the first two packets until the request times out, at which point that information would be discarded.

The problem here is that 1) waiting until *all* messages have been received before returning the data imposes an artificial delay in acting on it and 2) we don't need to receive the full response for the information to be useful.

## How can it be fixed?

These APIs should be updated to do roughly the same pattern here as is done with `NetworkAPI.recursive_find_nodes` which is to stream the responses asynchronously.

```python
# Client
async with client.find_nodes(...) as response_aiter:
    async for response in response_aiter:
        ...

# Network
async with network.find_nodes(...) as enrs_aiter:
    async for enr in enrs_aiter:
        ...
```

I think it would be best to look into implementing this as a *new* API as I think there is a time and place for both of these approaches.  Specifically, `NetworkAPI.lookup_enr` uses `await self.find_nodes(distance=0)` to retrieve the ENR record and using the `async with / async for` syntax in that case is overkill.  Suggested names for this new API:

- `ClientAPI.stream_find_nodes(...)`
- `NetworkAPI.stream_find_nodes(...)`

This will also apply to the new `ClientAPI.locate` introduced in #236

njgheorghita:
## What was wrong?

Fixes #255 

## How was it fixed?

Summary of approach.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/client.py
+++ b/ddht/v5_1/client.py
@@ -1,7 +1,8 @@
 import logging
 import socket
-from typing import Collection, List, Optional, Sequence, Tuple
+from typing import AsyncIterator, Collection, List, Optional, Sequence, Tuple, Type
 
+from async_generator import asynccontextmanager
 from async_service import Service
 from eth_enr import ENRAPI, ENRManager, QueryableENRDatabaseAPI
 from eth_keys import keys
@@ -20,8 +21,9 @@ from ddht.endpoint import Endpoint
 from ddht.enr import partition_enrs
 from ddht.message_registry import MessageTypeRegistry
 from ddht.request_tracker import RequestTracker
+from ddht.subscription_manager import SubscriptionManager
 from ddht.v5_1.abc import ClientAPI, EventsAPI
-from ddht.v5_1.constants import FOUND_NODES_MAX_PAYLOAD_SIZE
+from ddht.v5_1.constants import FOUND_NODES_MAX_PAYLOAD_SIZE, REQUEST_RESPONSE_TIMEOUT
 from ddht.v5_1.dispatcher import Dispatcher
 from ddht.v5_1.envelope import (
     EnvelopeDecoder,
@@ -31,6 +33,7 @@ from ddht.v5_1.envelope import (
 )
 from ddht.v5_1.events import Events
 from ddht.v5_1.messages import (
+    BaseMessage,
     FindNodeMessage,
     FoundNodesMessage,
     PingMessage,
@@ -418,6 +421,82 @@ class Client(Service, ClientAPI):
                     )
 
                 return responses
+
+    @asynccontextmanager
+    async def stream_find_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        distances: Collection[int],
+        *,
+        request_id: Optional[bytes] = None,
+    ) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[FoundNodesMessage]]]:
+        with self.request_tracker.reserve_request_id(node_id, request_id) as request_id:
+            request = AnyOutboundMessage(
+                FindNodeMessage(request_id, tuple(distances)), endpoint, node_id,
+            )
+
+            send_channel, receive_channel = trio.open_memory_channel[
+                InboundMessage[FoundNodesMessage]
+            ](256)
+            async with trio.open_nursery() as nursery:
+                nursery.start_soon(
+                    self._stream_find_nodes_response,
+                    request,
+                    request_id,
+                    FoundNodesMessage,
+                    send_channel,
+                )
+
+                try:
+                    async with receive_channel:
+                        try:
+                            yield receive_channel
+                        except trio.EndOfChannel as err:
+                            raise trio.TooSlowError from err
+                finally:
+                    nursery.cancel_scope.cancel()
+
+    async def _stream_find_nodes_response(
+        self,
+        request: AnyOutboundMessage,
+        request_id: Optional[bytes],
+        response_message_type: Type[BaseMessage],
+        send_channel: trio.abc.SendChannel[InboundMessage[FoundNodesMessage]],
+    ) -> None:
+        num_responses = 0
+
+        with trio.move_on_after(REQUEST_RESPONSE_TIMEOUT) as scope:
+            subscription_ctx = self.dispatcher.subscription_manager.subscribe(
+                response_message_type,
+                request.receiver_endpoint,
+                request.receiver_node_id,
+            )
+            async with subscription_ctx as subscription:
+                await self.dispatcher.send_message(request)
+
+                async with send_channel:
+                    async for response in subscription:
+                        if response.message.total == 0:
+                            raise ValidationError(
+                                f"Invalid `total` counter in response: total={response.message.total}"
+                            )
+
+                        if response.request_id != request_id:
+                            continue
+                        else:
+                            num_responses += 1
+                            await send_channel.send(response)
+
+                        if num_responses == response.message.total:
+                            break
+
+        if scope.cancelled_caught:
+            self.logger.debug(
+                "Stream find nodes request disconnected: request=%s message_type=%s",
+                request,
+                request_id,
+            )
 
     async def talk(
         self,
--- a/ddht/v5_1/client.py
+++ b/ddht/v5_1/client.py
@@ -498,9 +498,6 @@ async def common_client_stream_find_nodes(
     request_id: Optional[bytes] = None,
 ) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[FoundNodesMessage]]]:
     async def _stream_find_nodes_response(
-        client: Client,
-        request: AnyOutboundMessage,
-        request_id: Optional[bytes],
         response_message_type: Type[BaseMessage],
         send_channel: trio.abc.SendChannel[InboundMessage[FoundNodesMessage]],
     ) -> None:
@@ -510,23 +507,25 @@ async def common_client_stream_find_nodes(
 
         with trio.move_on_after(REQUEST_RESPONSE_TIMEOUT) as scope:
             async with send_channel:
-                subscription_ctx = client.dispatcher.subscribe(
-                    response_message_type,
-                    request.receiver_endpoint,
-                    request.receiver_node_id,
-                )
+                with client.request_tracker.reserve_request_id(
+                    node_id, request_id
+                ) as reserved_request_id:
+                    request = AnyOutboundMessage(
+                        FindNodeMessage(reserved_request_id, tuple(distances)),
+                        endpoint,
+                        node_id,
+                    )
 
-                async with subscription_ctx as subscription:
-                    async for response in subscription:
-                        if response.message.total == 0:
-                            raise ValidationError(
-                                "Invalid `total` counter in response: "
-                                f"total={response.message.total}"
-                            )
+                    async with client.dispatcher.subscribe_request(
+                        request, response_message_type
+                    ) as subscription:
+                        async for response in subscription:
+                            if response.message.total == 0:
+                                raise ValidationError(
+                                    "Invalid `total` counter in response: "
+                                    f"total={response.message.total}"
+                                )
 
-                        if response.request_id != request_id:
-                            continue
-                        else:
                             if response_counter == 0:
                                 head_response_total = response.message.total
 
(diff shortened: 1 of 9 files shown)


njgheorghita:
@pipermerriam I have a couple flaky tests & types left to clean up - but this pr is ready for a first pass :eyes: if you have some time for it today

ddht/v5_1/client.py
```diff
@@ -396,6 +399,82 @@ async def find_nodes(
...
+        request_id: Optional[bytes] = None,
+    ) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[FoundNodesMessage]]]:
+        with self.request_tracker.reserve_request_id(node_id, request_id) as request_id:
+            request = AnyOutboundMessage(
+                FindNodeMessage(request_id, tuple(distances)), endpoint, node_id,
+            )
+
+            send_channel, receive_channel = trio.open_memory_channel[
+                InboundMessage[FoundNodesMessage]
+            ](256)
```
pipermerriam, line 418: Given that we don't expect to regularly receive more than maybe 2-3 payloads for a given request in normal situations I would recommend lowering this limit to something like 4.

ddht/v5_1/client.py
```diff
@@ -396,6 +399,82 @@ async def find_nodes(
...
+                                f"Invalid `total` counter in response: total={response.message.total}"
+                            )
+
+                        if response.request_id != request_id:
+                            continue
+                        else:
+                            num_responses += 1
+                            await send_channel.send(response)
+
+                        if num_responses == response.message.total:
```
pipermerriam, line 468: I think this opens us up to potential paths for griefing.  Ideally, I think we should look at the first packet, pull the `total` value from that and use it as our *done* condition.  Additionally, we should *probably* enforce that all of the subsequent packets have the same `total` and throw a `ValidationError` if they don't.

ddht/v5_1/client.py
```diff
@@ -396,6 +399,82 @@ async def find_nodes(
...
+                    send_channel,
+                )
+
+                try:
+                    async with receive_channel:
+                        try:
+                            yield receive_channel
+                        except trio.EndOfChannel as err:
+                            raise trio.TooSlowError from err
+                finally:
```
pipermerriam, line 434: I'm not sure, but I think that this `try/finally` isn't needed.

1. If an exception occurs within the context of this function, it will trigger cancellation of the nursery.
2. If the context exits naturally, this will work just fine without the `try/finally`.

ddht/v5_1/client.py
```diff
@@ -396,6 +399,82 @@ async def find_nodes(
...
+                        if response.message.total == 0:
+                            raise ValidationError(
+                                f"Invalid `total` counter in response: total={response.message.total}"
+                            )
+
+                        if response.request_id != request_id:
+                            continue
+                        else:
+                            num_responses += 1
+                            await send_channel.send(response)
```
pipermerriam, line 466: I think this can raise `trio.BrokenResourceError` if the user exits the context before all responses have been returned.  You should be able to test this by setting up a request response where the response has multiple responses, but the receiver breaks out of the context after receiving the first response.

ddht/v5_1/network.py
```diff
@@ -403,21 +404,67 @@ async def find_nodes(
...
 
+    @asynccontextmanager
+    async def stream_find_nodes(
+        self,
+        node_id: NodeID,
+        endpoint: Endpoint,
+        distances: Collection[int],
+        *,
+        request_id: Optional[bytes] = None,
+    ) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[FoundNodesMessage]]]:
```
pipermerriam, line 419: This should be streaming the individual ENR records extracted from the responses you are streaming from the `ClientAPI.stream_find_nodes`

ddht/v5_1/network.py
```diff
@@ -403,21 +404,67 @@ async def find_nodes(
...
+        async with self.client.stream_find_nodes(
+            node_id, endpoint, distances=distances, request_id=request_id
+        ) as resp_aiter:
+            async with send_channel:
+                async for response in resp_aiter:
+                    # Validate that all responses are indeed at one of the
+                    # specified distances.
+                    validate_found_nodes_distances(response, node_id, distances)
+
+                    await send_channel.send(response)
```
pipermerriam, line 466: Here you should be sending the individual ENR records over the channel as the `NetworkAPI.stream_found_nodes` should stream the individual ENR records.

tests/core/v5_1/test_client.py
```diff
@@ -237,6 +240,138 @@ async def _send_response():
     assert len(checked_bucket_indexes) > 4
 
 
+@pytest.mark.trio
+async def test_client_request_response_stream_find_nodes_found_nodes(
+    alice, bob, alice_client, bob_client
+):
+    table = KademliaRoutingTable(bob.node_id, 256)
+    for i in range(1000):
```
pipermerriam, line 248: I think this test is probably overkill.  Here's what I think you want to make sure you are testing.

1. `ValidationError` case for returning ENR records with the wrong distance.
2. `ValidationError` case for `distance == 0` case
3. `ValidationError` case for `total == 0`
4. `ValidationError` for multiple responses with different `total` fields
5. That you can consume the full stream the full set of responses/ENR records
6. That you can consume only a few records out of a stream of response/ENR records and exit the context early
7. That we get a `TimeoutError` when the response messages don't show up

For these I think you can just create a few ENR records to populate the routing table and then request those specific distances.

ddht/v5_1/client.py
```diff
@@ -437,3 +463,126 @@ async def topic_query(
         request_id: Optional[bytes] = None,
     ) -> InboundMessage[FoundNodesMessage]:
         raise NotImplementedError
+
+
+@asynccontextmanager
+async def common_client_stream_find_nodes(
```
njgheorghita, line 469: This isn't actually "common" yet - I tried to implement it on the Alexandria client, but since an Alexandria client doesn't have a `dispatcher`, it wasn't a straightforward integration so I decided to leave it for a following pr if we want this api on the alexandria client/network

ddht/v5_1/client.py
```diff
@@ -437,3 +463,126 @@ async def topic_query(
...
+        ) as request_id:
+            request = AnyOutboundMessage(
+                FindNodeMessage(request_id, tuple(distances)), endpoint, node_id,
+            )
+
+            send_channel, receive_channel = trio.open_memory_channel[
+                InboundMessage[FoundNodesMessage]
+            ](4)
+            nursery.start_soon(
+                # mypy expects an Awaitable, interprets this as a Coroutine
```
njgheorghita, line 564: mypy interprets `_stream_find_nodes_response` as a `Coroutine` but expects an `Awaitable` - but from what I understand, `Coroutine` is a subclass of `Awaitable`. Any thoughts as to what could be problematic here?

ddht/v5_1/client.py
pipermerriam: I think you're right to do it in a separate PR.  Think it's going to be a little complicated to do it cleanly because we *can* expose a common `subscribe(...)` API but they use different message types for their `FindNodes` messages which makes a common implementation a bit more complex but doable.

First step is going to be proxying the subscription API so that it can be accessed directly from `NetworkAPI.subscribe(...)` (which under the hood simply proxies to the dispatcher).

Next you'll have to adjust this function to take the type of the `FindNodesMessage`.

And *then* you'll still have to deal with the fact that the messages are shaped differently and so you won't be able to access their payloads the same way... icky.

ddht/v5_1/client.py
pipermerriam: So I think what you're actually seeing is that `start_soon` doesn't like functions with more than 4 arguments.

```python
async def foo() -> None:
    async def arg_4(a: int, b: int, c: int, d: int) -> None:
        return
    async def arg_5(a: int, b: int, c: int, d: int, e: int) -> None:
        return

    async with trio.open_nursery() as nursery:
        nursery.start_soon(arg_4, 1, 2, 3, 4)
        nursery.start_soon(arg_5, 1, 2, 3, 4, 5)  # mypy gets mad about this.
```

You can work around this by simply using a `type: ignore`

ddht/v5_1/client.py
```diff
@@ -437,3 +463,111 @@ async def topic_query(
...
+async def common_client_stream_find_nodes(
+    client: Client,
+    node_id: NodeID,
+    endpoint: Endpoint,
+    distances: Collection[int],
+    *,
+    request_id: Optional[bytes] = None,
+) -> AsyncIterator[trio.abc.ReceiveChannel[InboundMessage[FoundNodesMessage]]]:
+    async def _stream_find_nodes_response(
+        client: Client,
```
pipermerriam, line 478: you can leave this out of the function args since it is available within the containing context.

pipermerriam (changes requested):
Right now you're using a combination of `dispatcher.send_message` and then the *generic* subscription API to get the responses (and then doing local filtering against the `request_id`).

I think you should instead use the `Dispatcher.subscribe_request` which lets you stream the responses and handles all of this for you.

ddht/v5_1/client.py
njgheorghita: Ah, yes of course and this also takes care of the mypy complaining about too many arguments for `nursery.start_soon()`!

ddht/v5_1/client.py
```diff
@@ -460,3 +486,93 @@ async def topic_query(
...
+                    ) as subscription:
+                        async for response in subscription:
+                            if response.message.total == 0:
+                                raise ValidationError(
+                                    "Invalid `total` counter in response: "
+                                    f"total={response.message.total}"
+                                )
+
+                            if response_counter == 0:
+                                head_response_total = response.message.total
```
pipermerriam, line 530: I'm curious what you think about the following pattern.

```python
head_response = await subscription.receive()
yield head_response

total_expected_messages = head_response.message.total

for _ in range(total_expected_messages - 1):  # since we already received one...
    response = await subscription.receive()
    yield response
```

I like that it gets rid of this logic for detecting that it is our first pass through the loop.  I think it trades that for a little bit of awkwardness of not actually using the channel as an async iterator which is a very natural way to do things and requires instead that we explicitely call `receive()`.  Thoughts?  I'm not convinced the structure above is better.

ddht/v5_1/network.py
```diff
@@ -540,6 +541,53 @@ async def _monitor_done(
...
+            async with send_channel:
+                async for response in resp_aiter:
+                    enrs = response.message.enrs
+                    for enr in enrs:
+                        try:
+                            await send_channel.send(enr)
+                        except (trio.BrokenResourceError, trio.ClosedResourceError):
+                            break
+
+    if not distances:
```
pipermerriam, line 568: *nitpick* maybe move this above the function definition since we typically want to error out as early as possible.

ddht/v5_1/network.py
```diff
@@ -540,6 +541,53 @@ async def _monitor_done(
...
+        raise TypeError("Must provide at least one distance")
+
+    if endpoint is None:
+        endpoint = await network.endpoint_for_node_id(node_id)
```
pipermerriam, line 572: I think you should move this inside of the `_stream_find_nodes_response` it took me a minute to figure out what was going on as this variable *appears* to not be used but you have to scroll back up to the function definition to find it being used inside the context.

ddht/validation.py
```diff
@@ -136,3 +137,20 @@ def validate_and_extract_destination(value: Any) -> Tuple[NodeID, Optional[Endpo
         raise RPCError(f"Unrecognized node identifier: {value}")
 
     return node_id, endpoint
+
+
+def validate_found_nodes_distances(
```
pipermerriam, line 142: Maybe write a test for this helper.

ddht/validation.py
```diff
@@ -136,3 +137,20 @@ def validate_and_extract_destination(value: Any) -> Tuple[NodeID, Optional[Endpo
...
+    enrs: Collection[ENRAPI], node_id: NodeID, distances: Collection[int],
```
pipermerriam, line 143: `local_node_id` might be a more appropriate variable name here.

pipermerriam (approved):
Solid.   Follow up work once this is merged is to update `common_recursive_find_nodes` to leverage this API (but you'll have to do the integration into Alexandria first).

pipermerriam (commented):
I just realized that once you have these working you should probably re-work the non-streaming versions to use the streaming versions under the hood.  Should remove a bunch of loosely duplicate code.

```python
class Network:
    async def find_nodes(...):
        async with self.stream_find_nodes(...) as something_aiter:
            return tuple([value async for value in something_aiter])
```

ddht/v5_1/client.py
njgheorghita: It's not obvious to me that one option's better than the other. But I prefer removing the unnecessarily run logic over whatever slight gain in readability from using the channel "properly" (since imo it's still very obvious as to what's happening)

njgheorghita: Now that I'm tinkering with it - we're not just `yield response`ing here, but sending the response over the `send_channel`, which comes with it's own `try/except` and doing some validation, so in this case it seems a little to cumbersome duplicating all that logic to make whatever slight gain in performance worth it.

pipermerriam: Here's my implementation of roughly the same logic for the Alexandria APIs that have roughly the same shape:

https://github.com/ethereum/ddht/pull/278/files

I didn't end up putting any `try/except` around the channel operations.  I'm not 100% sure they aren't needed, but I *think* that it might not be necessary.  I'd at least look into removing the handling and see if you get errors in a few test runs locally and in CI.  If nothing blows up :+1:  and if it blows up later we know where to add the handling.

njgheorghita:
@pipermerriam Sure thing, the only difference being that we want to support distances of `0` for `find_nodes()` and not `stream_find_nodes()`, correct? i.e. you can request a remote to send it's own enr via `find_nodes()` but not `stream_find_nodes()`?

pipermerriam:
I don't think we need to restrict `stream_find_nodes` since we can do queries that include distance `0` **and** other distances.

PR #264, opened 2020-12-07
Handle ConnectionResetError in rpc server

pipermerriam:
## What was wrong?

A `ConnectionResetError` was causing flakey tests and the crash of the `RPCServer`

## How was it fixed?

Simple error handling.

#### Cute Animal Picture

Hare-660x330


pipermerriam:
Tests got really flaking for a bit recently.  I think I'm making headway moving the needle the other direction.

PR #267, opened 2020-12-09
move ResourceQueue

pipermerriam:
## What was wrong?

The `ResourceQueue` doesn't need to be in the alexandria module

## How was it fixed?

Moved it to the base of the library.

#### Cute Animal Picture

penguin-jumping


ddht/resource_queue.py
```diff
@@ -1,16 +1,20 @@
...
 
 class RemoveResource(Exception):
     pass
 
 
-class ResourceQueue(Container[TResource]):
+class EmptyResourceQueue(Exception):
+    pass
+
+
```
pipermerriam, line 16:
```suggestion
```

PR #278, opened 2020-12-09
Implement location streaming APIs for Alexandria

pipermerriam:
## What was wrong?

When searching for content we need to be able to quickly find advertisements.

## How was it fixed?

Leverages multiple layers of async streaming APIs that allow us to quickly zero in on a part of the network (`recursive_find_nodes(...)`) and then enumerate the nodes in that area of the network (`explore(...)`, and then query them for advertisements  `stream_locate(...)`.

#### Cute Animal Picture

Funny Animals Using Mobiles (1)

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -539,30 +539,49 @@ class AlexandriaClient(Service, AlexandriaClientAPI):
         content_key: ContentKey,
         request_id: Optional[bytes] = None,
     ) -> Tuple[InboundMessage[LocationsMessage], ...]:
+        stream_locate_ctx = self.stream_locate(
+            node_id, endpoint, content_key=content_key, request_id=request_id,
+        )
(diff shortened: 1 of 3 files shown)


ddht/v5_1/alexandria/client.py
```diff
@@ -539,30 +539,49 @@ async def locate(
         content_key: ContentKey,
         request_id: Optional[bytes] = None,
     ) -> Tuple[InboundMessage[LocationsMessage], ...]:
+        stream_locate_ctx = self.stream_locate(
+            node_id, endpoint, content_key=content_key, request_id=request_id,
+        )
+        async with stream_locate_ctx as response_aiter:
+            return tuple([response async for response in response_aiter])
```
pipermerriam, line 546: @njgheorghita here's a loose equivalent to what you're doing where I was able to implement the streaming version and then move the non-streaming version over to just depend on the streaming one.

PR #279, opened 2020-12-10
Implement event loop canary

pipermerriam:
## What was wrong?

We want to know if the event loop is blocked.

## How was it fixed?

Run a service that just loops and checks for unexpected delays.

#### Cute Animal Picture

DogCatPhone

### DIFF
--- a/ddht/canary.py
+++ b/ddht/canary.py
@@ -1,0 +1,36 @@
+import logging
+
+from async_service import Service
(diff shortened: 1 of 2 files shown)


pipermerriam:
Well, we're at least starting from a good spot.  At least running on my machine we're seeing delays averaging `0.001` second which is effectively no delay.

Issue #172, opened 2020-11-09
Restrict Alexandria messages to appropriate TALKREQ or TALKRESP


pipermerriam:
## What was wrong?

Currently, it *may* be possible to send a `PingMessage` using `TALKRESP` or a `PongMessage` using `TALKREQ` in Alexandria.  All messages should be flagged as request or response and the protocol should enforce this.

## How can it be fixed?

Another class level field should be placed on the message classes indicating that they are of type `Request` or `Response`.  We can then refuse to decode messages that are coming in from the wrong base message type.

pipermerriam:
@gsalgado this one is reasonably simple/small

PR #308, opened 2020-12-14
Restrict Alexandria messages to appropriate TALKREQ or TALKRESP

gsalgado:
Closes: #172

### DIFF
--- a/ddht/v5_1/alexandria/client.py
+++ b/ddht/v5_1/alexandria/client.py
@@ -35,6 +35,7 @@ from ddht.v5_1.alexandria.messages import (
     AckMessage,
     AdvertiseMessage,
     AlexandriaMessage,
+    AlexandriaMessageType,
     ContentMessage,
     FindNodesMessage,
     FoundNodesMessage,
@@ -120,6 +121,8 @@ class AlexandriaClient(Service, AlexandriaClientAPI):
         *,
         request_id: bytes,
     ) -> None:
+        if message.type != AlexandriaMessageType.RESPONSE:
+            raise TypeError("%s is not of type RESPONSE", message)
         data_payload = message.to_wire_bytes()
         await self.network.client.send_talk_response(
             node_id, endpoint, payload=data_payload, request_id=request_id,
@@ -155,6 +158,8 @@ class AlexandriaClient(Service, AlexandriaClientAPI):
         # then feed this into our local tracker, which allows us to query it
         # upon receiving an incoming TALKRESPONSE to see if the response is to
         # a message from this protocol.
+        if request.type != AlexandriaMessageType.REQUEST:
+            raise TypeError("%s is not of type REQUEST", request)
         if request_id is None:
             request_id = self.network.client.request_tracker.get_free_request_id(
                 node_id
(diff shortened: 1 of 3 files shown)


ddht/v5_1/alexandria/client.py
```diff
@@ -598,6 +603,10 @@ async def _feed_talk_requests(self) -> None:
                 except DecodingError:
                     pass
                 else:
+                    if message.type != AlexandriaMessageType.REQUEST:
+                        self.logger.warning("Received non-REQUEST msg via TALKREQ: %s", message)
```
gsalgado, line 607: In the issue description you mentioned avoiding the decoding if the msg is of the wrong type, but I couldn't find a way to achieve that since the type is included in the underlying msg's payload.

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -48,6 +49,60 @@ async def _respond_empty():
...
+        await alice_alexandria_client._send_response(
+            bob.node_id,
+            bob.endpoint,
+            PingMessage(PingPayload(enr_seq=1234, advertisement_radius=4321)),
+            request_id=b"\x01\x02",
+        )
+
+
+@pytest.mark.trio
+async def test_feed_talk_requests_ignores_wrongly_typed_msgs(
```
gsalgado, line 76: We're missing a similar test for `_feed_talk_responses`. I'll add one if you're happy with the approach I used here to fix the issue

ddht/v5_1/alexandria/client.py
```diff
@@ -625,6 +636,12 @@ async def _feed_talk_responses(self) -> None:
                 except DecodingError:
                     pass
                 else:
+                    if message.type != AlexandriaMessageType.RESPONSE:
+                        self.logger.warning(
```
pipermerriam, line 640: I think this should be `DEBUG`.  Otherwise, a malfunctioning (or malicious) client will cause log spam.

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -48,6 +49,63 @@ async def _respond_empty():
...
+            # The message will be received by bob.
+            with trio.fail_after(1):
+                inbound_msg = await sub.receive()
+
+            assert inbound_msg.request_id == b"\x01\x02"
+
+            # But the alexandria client will not deliver it to any subscribers. And it should
+            # log a warning about it.
+            pong_msg = None
+            with trio.move_on_after(0.2):
```
pipermerriam, line 103: You can raise this timeout a little to avoid false positives and use the [`autojump_clock`](https://pytest-trio.readthedocs.io/en/latest/quickstart.html#trio-s-magic-autojump-clock) to avoid having to actually wait for the timeout.

tests/core/v5_1/alexandria/test_alexandria_client.py
```diff
@@ -48,6 +49,63 @@ async def _respond_empty():
...
+                pong_msg = await client_subscription.receive()
+
+            assert pong_msg is None
```
pipermerriam, line 106: Instead of this you can do:

```python
with trio.move_on_after(2) as scope:
    ...

assert scope.cancelled_caught
```

pipermerriam (approved):

PR #338, opened 2020-12-21
Add lru sql dict tooling

### LINKED ISSUE #283: SQLite backed key/value store for BroadcastLog
## What was wrong?

The `BroadcastLog._log: Dict[Hash32, int]` stores the last time an advertisement was sent to a certain node.  Currently the class is limited at 8192 which is far too small for real world scenarios.  Supposing we have a million advertisements and we send them each to 10 nodes, this store will be ~500MB.

## How can it be fixed?

Write a small `sqlite3` backed `Dict` that we can use.  It will need a cleanup mechanism of some sort so this should potentially be written as an LRU.

njgheorghita:
## What was wrong?
Fixes #283 

## How was it fixed?
Update `BroadcastLog` to use a sql-backed LRU dict.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/tools/lru_sql_dict.py
+++ b/ddht/tools/lru_sql_dict.py
@@ -1,0 +1,271 @@
+import pickle
+import sqlite3
+from typing import Any, Iterator, MutableMapping, NamedTuple, Tuple, Union
+
+from eth_utils.toolz import first
+
+CREATE_CACHE_QUERY = """
+    CREATE TABLE IF NOT EXISTS cache (
+        id INTEGER NOT NULL PRIMARY KEY,
+        key BLOB NOT NULL UNIQUE,
+        value BLOB NOT NULL,
+        pref INTEGER UNIQUE,
+        nref INTEGER UNIQUE
+    )
+"""
+
+CREATE_MAP_QUERY = """
+    CREATE TABLE IF NOT EXISTS map (
+        key BLOB NOT NULL PRIMARY KEY,
+        pointer INTEGER NOT NULL UNIQUE
+    )
+"""
+
+DEFAULT_CACHE_SIZE = 100000
+
+
+class Node(NamedTuple):
+    id: int
+    key: Union[str, bytes]
+    value: Any
+    pref: int
+    nref: int
+
+
+class LRUSQLDict(MutableMapping[Union[str, bytes], Any]):
+    """
+    SQLite3-backed dictionary that implements an LRU cache.
+    """
+
+    def __init__(
+        self, conn: sqlite3.Connection, cache_size: int = DEFAULT_CACHE_SIZE
+    ) -> None:
+        self._conn = conn
+        self._execute(CREATE_CACHE_QUERY)
+        self._execute(CREATE_MAP_QUERY)
+        self.cache_size = cache_size
+
+    def __iter__(self) -> Iterator[Union[str, bytes]]:
+        with self._conn:
+            for key in self._conn.execute("SELECT key FROM cache"):
+                yield first(key)
+
+    def __len__(self) -> int:
+        result = self._fetch_single_query("SELECT COUNT(*) FROM cache;")
+        return int(first(result))
+
+    def __setitem__(self, key: Union[str, bytes], value: Any) -> None:
+        # updates LRU cache
+        try:
+            self.__getitem__(key)
+        except KeyError:
+            self._insert_item(key, value)
+        else:
+            self._update_item(key, value)
+
+    def __getitem__(self, key: Union[str, bytes]) -> Any:
+        # updates LRU cache
+        pointer_result = self._fetch_single_query(
+            "SELECT pointer FROM map WHERE key=?;", (key,),
+        )
+
+        if not pointer_result:
+            raise KeyError(f"Key ({str(key)}) not found in db.")
+
+        (map_pointer,) = pointer_result
+
+        lookup_result = self._fetch_single_query(
+            "SELECT value FROM cache WHERE id=?;", (map_pointer,),
+        )
+
+        (value,) = lookup_result
+
+        unpickled_value = self._unpickle(value)
+
+        # update cache
+        self.__delitem__(key)
+        self._insert_item(key, unpickled_value)
+
+        return unpickled_value
+
+    def __delitem__(self, key: Union[str, bytes]) -> None:
+        result = self._fetch_single_query(
+            "SELECT id,key,value,pref,nref FROM cache WHERE key=?;", (key,),
+        )
+
+        if not result:
+            raise KeyError(f"Cannot delete key. Key: {str(key)} not found in db.")
+
+        node = Node(*result)
+
+        # delete key from cache & map
+        self._execute(
+            "DELETE FROM map WHERE pointer=?;", (node.id,),
+        )
+        self._execute(
+            "DELETE FROM cache WHERE id=?;", (node.id,),
+        )
+
+        # update any nrefs/prefs in cache
+        nref_node = self._fetch_single_query(
+            "SELECT id,key,value,pref,nref FROM cache WHERE nref=?;", (node.id,),
+        )
+        pref_node = self._fetch_single_query(
+            "SELECT id,key,value,pref,nref FROM cache WHERE pref=?;", (node.id,),
+        )
+
+        if nref_node and pref_node:
+            nref = Node(*nref_node)
+            pref = Node(*pref_node)
+            self._execute(
+                "UPDATE cache SET nref=? WHERE id=?;", (pref.id, nref.id),
+            )
+            self._execute(
+                "UPDATE cache SET pref=? WHERE id=?;", (nref.id, pref.id),
+            )
+        elif nref_node:
+            nref = Node(*nref_node)
+            self._execute(
+                "UPDATE cache SET nref=? WHERE id=?;", (None, nref.id),
+            )
+        elif pref_node:
+            pref = Node(*pref_node)
+            self._execute(
+                "UPDATE cache SET pref=? WHERE id=?;", (None, pref.id),
+            )
+
+    def _insert_item(self, key: Union[str, bytes], value: Any) -> None:
+        # insert new item into map and cache
+        pickled_value = self._pickle(value)
+        if self.is_empty:
+            new_pref = None
+            new_nref = None
+            new_pkey = 0
+
+            self._execute(
+                "INSERT INTO cache VALUES (?, ?, ?, ?, ?);",
+                (new_pkey, key, pickled_value, new_pref, new_nref),
+            )
+
+            self._execute("INSERT INTO map VALUES (?, ?);", (key, new_pkey))
+
+        else:
+            # evict least recently used key/value
+            if self.is_full:
+                self.__delitem__(self.tail.key)
+
+            # get old head
+            old_head_pkey, _, _, old_head_pref, _ = self.head
+
+            # get attributes for new head
+            (max_pkey,) = self._fetch_single_query(
+                "SELECT id FROM cache ORDER BY id DESC LIMIT 1;"
+            )
+            new_pkey = max_pkey + 1
+            new_pref = None
+            new_nref = old_head_pkey
+
+            # add new head to cache
+            self._execute(
+                "INSERT INTO cache VALUES (?, ?, ?, ?, ?);",
+                (new_pkey, key, pickled_value, new_pref, new_nref),
+            )
+
+            # add new head to map
+            self._execute(
+                "INSERT INTO map VALUES (?, ?);", (key, new_pkey),
+            )
+
+            # update old head in cache
+            self._execute(
+                "UPDATE cache SET pref=? WHERE id=?;", (new_pkey, old_head_pkey),
+            )
+
+    def _update_item(self, key: Union[str, bytes], value: Any) -> None:
+        pickled_value = self._pickle(value)
+        self._execute("UPDATE cache SET value=? WHERE key=?;", (pickled_value, key))
+
+    def _fetch_single_query(self, query: str, args: Tuple[Any, ...] = None) -> Any:
+        with self._conn:
+            if args:
+                cursor = self._conn.execute(query, args).fetchall()
+            else:
+                cursor = self._conn.execute(query).fetchall()
+        if len(cursor) > 1:
+            raise Exception(
+                f"Invalid db state. More than one result found for query: {query}."
+            )
+        if not cursor:
+            return None
+        return first(cursor)
+
+    def _execute(self, query: str, args: Tuple[Any, ...] = None) -> None:
+        with self._conn:
+            if args:
+                self._conn.execute(query, args)
+            else:
+                self._conn.execute(query)
+
+    def _pickle(self, value: Any) -> bytes:
+        return sqlite3.Binary(pickle.dumps(value, -1))
+
+    def _unpickle(self, value: bytes) -> Any:
+        return pickle.loads(value)
+
+    @property
+    def is_full(self) -> bool:
+        return self.__len__() >= self.cache_size
+
+    @property
+    def is_empty(self) -> bool:
+        return self.__len__() == 0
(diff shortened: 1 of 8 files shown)


ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
+import pickle
+import sqlite3
+from typing import Any, Iterator, MutableMapping, NamedTuple, Tuple, Union
+
+from eth_utils.toolz import first
+
+CREATE_CACHE_QUERY = """
+    CREATE TABLE IF NOT EXISTS cache (
+        id INTEGER NOT NULL PRIMARY KEY,
+        key BLOB NOT NULL UNIQUE,
```
pipermerriam, line 10: Any reason why you didn't make this the primary key?

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+    nref: int
+
+
+class LRUSQLDict(MutableMapping[Union[str, bytes], Any]):
+    """
+    SQLite3-backed dictionary that implements an LRU cache.
+    """
+
+    def __init__(
+        self, conn: sqlite3.Connection, cache_size: int = DEFAULT_CACHE_SIZE
```
pipermerriam, line 41: Two thoughts.

1. I don't think this should be have default size
2. Allowing `None` as the default size would allow for using this without bounds... but we don't necessarily have an immediate use case for that.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+        self.cache_size = cache_size
+
+    def __iter__(self) -> Iterator[Union[str, bytes]]:
+        with self._conn:
+            for key in self._conn.execute("SELECT key FROM cache"):
+                yield first(key)
+
+    def __len__(self) -> int:
+        result = self._fetch_single_query("SELECT COUNT(*) FROM cache;")
+        return int(first(result))
```
pipermerriam, line 55: I assume the `int` cast is to make `mypy` happy?  If so I think a `# type: ignore` comment would be more appropriate to not incur any runtime overhead (albeit a very small amount).

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+            self._update_item(key, value)
+
+    def __getitem__(self, key: Union[str, bytes]) -> Any:
+        # updates LRU cache
+        pointer_result = self._fetch_single_query(
+            "SELECT pointer FROM map WHERE key=?;", (key,),
+        )
+
+        if not pointer_result:
+            raise KeyError(f"Key ({str(key)}) not found in db.")
```
pipermerriam, line 73: *nitpick*

Convention for `KeyError` is to just have the key itself be the only argument to the exception class.  This only really matters in cases where the entity catching the exception wants to extract the key from the exception which is probably not the greatest idea since there's no guarantees that the entity raising the exception is following this convention....

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def _execute(self, query: str, args: Tuple[Any, ...] = None) -> None:
+        with self._conn:
+            if args:
+                self._conn.execute(query, args)
+            else:
+                self._conn.execute(query)
+
+    def _pickle(self, value: Any) -> bytes:
+        return sqlite3.Binary(pickle.dumps(value, -1))
```
pipermerriam, line 210: Can we allow for custom serializers to avoid the need for `pickle` which tends to be slow.

For most values we want to store we can do the serialization faster and just as easy.  This probably means requiring a *serializer* to be passed in which could be as simple as:

```python
TKey = TypeVar("TKey")
TValue = TypeVar("TValue")

class LRUSQLDict(MutableMapping[TKey, TMapping]):
    def __init__(self,
                 key_encoder: Callable[[TKey], bytes],
                 key_decoder: Callable[[bytes], TKey],
                 value_encoder: Callable[[TValue], bytes],
                 value_decoder: Callable[[bytes], TValue],
                 ...) -> None:
        ...
```

It's a bit verbose, but they are easy functions to provide and avoiding pickle, especially for things persisted in a database that are long lived is going to save us some headache.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+    @property
+    def tail(self) -> Node:
+        tail = self._fetch_single_query(
+            "SELECT id,key,value,pref,nref FROM cache WHERE nref IS NULL;"
+        )
+        if not tail:
+            raise Exception("No tail found.")
+        return Node(*tail)
+
+    # custom iterator type not compatible with supertype
```
pipermerriam, line 241: This seems like it should be something you can satisfy.  You probably need to update this class to use [generic typing](https://docs.python.org/3/library/typing.html#user-defined-generic-types)

pipermerriam (commented):
Can you add a comment explaining the SQL schema?

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,271 @@
...
+        )
+        pref_node = self._fetch_single_query(
+            "SELECT id,key,value,pref,nref FROM cache WHERE pref=?;", (node.id,),
+        )
+
+        if nref_node and pref_node:
+            nref = Node(*nref_node)
+            pref = Node(*pref_node)
+            self._execute(
+                "UPDATE cache SET nref=? WHERE id=?;", (pref.id, nref.id),
```
pipermerriam, line 121: You should be able to batch all of these updates using [`Connection.executemany`](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.executemany)

ddht/tools/lru_sql_dict.py
njgheorghita: Not particularly, it seemed like best practice from what I read for an lru cache to have a hashmap that points to a doubly linked list - which is what I was going for here. But it does seem doable to remove the hashmap entirely and just have a single table that acts as the cache w/ `key (primary key), value, pref, nref`.

ddht/tools/lru_sql_dict.py
njgheorghita: I've updated the default size to be `None`. My initial intent with the default size was to avoid a db-backed dict from being created with some cache-size, and then later on that same db being instantiated with a smaller cache size - messing up the cache integrity. Is this a case to be worried about / handled?

ddht/tools/lru_sql_dict.py
njgheorghita: I'm not sure it's the right use case here - since the query is slightly different in each case, and from what I understand with `executemany`, you want to provide a sequence of args against the same query.

ddht/tools/lru_sql_dict.py
njgheorghita: I'm having a little trouble with this one. I've updated the types (for `items()` and `values()`) to be compatible with the supertype's signature, which is defined with `ValueViews` and `AbstractSet` - but then mypy complains that it is expecting a `Generator` type due to the `yield`. I haven't been able to find some way to reconcile these different types, and it's not obvious to me how I could achieve this using generics.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def __iter__(self) -> Iterator[TKey]:
+        with self._conn:
+            for key in self._conn.execute("SELECT key FROM cache"):
+                yield self.key_decoder(first(key))
+
+    def __len__(self) -> int:
+        result = self._fetch_single_query("SELECT COUNT(*) FROM cache;")
+        # ignore b/c mypy cannot interpret the result as an integer
+        return first(result)  # type: ignore
```
pipermerriam, line 76: *nitpick*

suppose that `result` had more than one item.  This would silently leave that second item behind.

```python
(result,) = self._fetch_single_query("SELECT COUNT(*) FROM cache;") 
```

This unpacks the result, which will throw an exception if there is more than one item.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+        value_encoder: Callable[[TValue], bytes],
+        value_decoder: Callable[[bytes], TValue],
+        cache_size: int = None,
+    ) -> None:
+        self._conn = conn
+        self.cache_size = cache_size
+        self.key_encoder = key_encoder
+        self.key_decoder = key_decoder
+        self.value_encoder = value_encoder
+        self.value_decoder = value_decoder
```
pipermerriam, line 64: *nitpick*  Maybe make these private since I don't think it's appropriate for an external consumer to access them.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+            for key in self._conn.execute("SELECT key FROM cache"):
+                yield self.key_decoder(first(key))
+
+    def __len__(self) -> int:
+        result = self._fetch_single_query("SELECT COUNT(*) FROM cache;")
+        # ignore b/c mypy cannot interpret the result as an integer
+        return first(result)  # type: ignore
+
+    def __setitem__(self, key: TKey, value: TValue) -> None:
+        # updates LRU cache
```
pipermerriam, line 79: I understand what you are saying here but I'm not sure a casual reader would.  It might be worthwhile to be a bit more verbose here and explain that accessing the item resets it to the front of the cache.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def __getitem__(self, key: TKey) -> TValue:
+        # updates LRU cache
+        serialized_key = self.key_encoder(key)
+        lookup_result = self._fetch_single_query(
+            "SELECT value FROM cache WHERE key=?;", (serialized_key,),
+        )
+
+        if not lookup_result:
+            raise KeyError(str(key))
```
pipermerriam, line 95: I don't think you need the `str` casting here.  If the key is `bytes` or an integer, it seems ok to just keep it as that type.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+
+        (value,) = lookup_result
+
+        deserialized_value = self.value_decoder(value)
+
+        # update cache
+        self.__delitem__(key)
```
pipermerriam, line 102: *nitpick*

It seems unfortunate to fully delete the row and re-insert.  My understanding is that what we want to do is update the surrounding rows in the doubly linked list to point to each other, and then to find the row head row, point it at this record, and update this record so that it's `nref` points at `None`.

I can see how modeling this as delete -> insert accomplishes the same thing, but it would be nice to not do the extra work.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+        self._execute(
+            "DELETE FROM cache WHERE key=?;", (node.key,),
+        )
+
+        # update any nrefs/prefs in cache
+        nref_node = self._fetch_single_query(
+            "SELECT key,value,pref,nref FROM cache WHERE nref=?;", (node.key,),
+        )
+        pref_node = self._fetch_single_query(
+            "SELECT key,value,pref,nref FROM cache WHERE pref=?;", (node.key,),
```
pipermerriam, line 128: Looks like you are selecting more than you need to here since you only access the `key` part of the node.  Can you update these queries to only select `key`?

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+            new_nref = None
+
+            self._execute(
+                "INSERT INTO cache VALUES (?, ?, ?, ?);",
+                (serialized_key, serialized_value, new_pref, new_nref),
+            )
+
+        else:
+            # evict least recently used key/value
+            if self.is_full:
```
pipermerriam, line 166: If you changed this to `while self.is_full:` then it seems like it would account for the situation where someone re-uses the same database with a different max size.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+            )
+
+    def _update_item(self, key: TKey, value: TValue) -> None:
+        serialized_key = self.key_encoder(key)
+        serialized_value = self.value_encoder(value)
+        self._execute(
+            "UPDATE cache SET value=? WHERE key=?;", (serialized_value, serialized_key)
+        )
+
+    def _fetch_single_query(self, query: str, args: Tuple[Any, ...] = None) -> Any:
```
pipermerriam, line 190: Since tuples are immutable, you can *in theory* change this to just be:

```
     def _fetch_single_query(self, query: str, args: Tuple[Any, ...] = ()) -> Any: 
```

This should let you remove the `if/else` below since it appears that `conn.execute` is happy to have an empty tuple passed in for `args` when the query has no args.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+                cursor = self._conn.execute(query).fetchall()
+        if len(cursor) > 1:
+            raise Exception(
+                f"Invalid db state. More than one result found for query: {query}."
+            )
+        if not cursor:
+            return None
+        return first(cursor)
+
+    def _execute(self, query: str, args: Tuple[Any, ...] = None) -> None:
```
pipermerriam, line 204: Same here.  I think you can change the default of `args` to the empty tuple.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+    def is_empty(self) -> bool:
+        return self.__len__() == 0
+
+    @property
+    def head(self) -> Node:
+        head = self._fetch_single_query(
+            "SELECT key,value,pref,nref FROM cache WHERE pref IS NULL;"
+        )
+        if not head:
+            raise Exception("No head found.")
```
pipermerriam, line 227: I'm not 100% on this being the right behavior, but `KeyError` or `IndexError` seem appropriate.  Conceptually this is close to `dict.pop()`.

ddht/tools/lru_sql_dict.py
pipermerriam: I'm not sure it is worth doing but:

```python
In [41]: collections.abc.ValuesView({'a': 1, 'b': 2})
Out[41]: ValuesView({'a': 1, 'b': 2})

In [42]: tuple(_41)
Out[42]: (1, 2)
```

It looks like you can instantiate your own `ValuesView` but it also looks like it'd be quite awkward to make this work with that API....  So :+1: on just `# type: ignore` but it would be cool if there was a clean way to do this.

ddht/tools/lru_sql_dict.py
```diff
@@ -0,0 +1,269 @@
...
+    # custom iterator type not compatible with supertype
+    def values(self) -> ValuesView[TValue]:  # type: ignore
+        for key in self.__iter__():
+            result = self._fetch_single_query(
+                "SELECT value FROM cache WHERE key=?;", (key,)
+            )
+            yield self.value_decoder(first(result))
+
+    # custom iterator type not compatible with supertype
+    def items(self) -> AbstractSet[Tuple[TKey, TValue]]:  # type: ignore
```
pipermerriam, line 248: I think this should be `ItemsView`?

pipermerriam (approved):
:+1: 

If you have any concerns over testing, consider writing a fuzz test that performs the same operations on both your implementation and `lru.LRU`.  You can randomly insert/get/delete/etc and simply assert that they each behave the same at each stage.

ddht/tools/lru_sql_dict.py
njgheorghita: I agree that it's not the most efficient implementation, but it removes a fair amount of duplicated code and handles some edge cases in a tidy manner - at the cost of (imo) only a slight amount of extra overhead. I left it as a `todo` to find a better implementation - that I can revisit in a follow-up pr if you're worried about the performance.

PR #340, opened 2021-03-02
POC - Simple Key/Value DHT

pipermerriam:
WIP for state availability testnets


pipermerriam:
@carver still missing tests for most of the content retrieval APIs and the JSON-RPC APIs for interacting with content.

At present this:

1. implements a single content data store (currently only in-memory, but trivial to migrate to an on-disk sqlitedb)
2. listens for and serves content requests from that data store
3. expores JSON-RPC APIs for: get/set/delete local content as well as retrieval of content by finding it on the network.

I think it's worth adding tests since I did a bunch of manual futzing to get this working and we'll want to be able to actually iterate on this with knowledge that it remains functional.  Otherwise, this branch implements everything we need to spin up a test network.

ddht/boot_info.py
```diff
@@ -55,6 +55,8 @@ def _cli_args_to_boot_info_kwargs(args: argparse.Namespace) -> BootInfoKwargs:
     if args.listen_address is None:
         listen_on = None
     else:
+        import ipaddress
+        assert isinstance(args.listen_address, ipaddress.IPv4Address)
```
pipermerriam, line 59: debug code that needs to be removed.

ddht/v5_1/alexandria/seeker.py
```diff
@@ -0,0 +1,137 @@
...
+        self.content_id = content_key_to_content_id(content_key)
+
+        self._network = network
+
+        self._content_send, self.content_receive = trio.open_memory_channel[bytes](0)
+
+    async def run(self) -> None:
+        async with self.content_receive, self._content_send:
+            raw_enr_send, raw_enr_receive = trio.open_memory_channel[ENRAPI](32)
+            sorted_enr_send, sorted_enr_receive = trio.open_memory_channel[ENRAPI](0)
```
carver, line 43: It's not clear on first read how `raw` and `sorted` are being used here. I'll come back with a suggestion as soon as I understand it. :)

... Maybe `closest_` and `recent_`?

ddht/v5_1/alexandria/seeker.py
```diff
@@ -0,0 +1,137 @@
...
+
+            self.manager.run_daemon_task(self._explore_for_enrs, raw_enr_send)
+            self.manager.run_daemon_task(
+                self._collate, sorted_enr_send, raw_enr_receive
+            )
+
+            for _ in range(self.concurrency):
+                self.manager.run_daemon_task(
+                    self._worker, raw_enr_send, sorted_enr_receive,
```
carver, line 52: Yeah, I think I might have breezed by this method if it had been something like:
```suggestion
                    self._seek, recent_enr_send, closest_enr_receive,
```

ddht/v5_1/app.py
```diff
@@ -137,7 +137,7 @@ async def run(self) -> None:
         self.logger.info(
             "Local ENR       : seq=%d enr=%s",
             enr_manager.enr.sequence_number,
-            enr_manager.enr,
+            repr(enr_manager.enr),
```
carver, line 140: I think it's cleaner to leave this unchanged and switch the log to `%r`

ddht/v5_1/constants.py
```diff
@@ -28,13 +28,23 @@
...
+
 PACKET_VERSION_1 = b"\x00\x01"
 
 ID_NONCE_SIGNATURE_PREFIX = b"discovery v5 identity proof"
 
 HEADER_PACKET_SIZE = 23
 
-PROTOCOL_ID = b"discv5"
+# PROTOCOL_ID = b"discv5"
+# XXX: Must be the same length as "discv5"
```
carver, line 46: Interesting, a discv5 requirement?

ddht/v5_1/constants.py
```diff
@@ -28,13 +28,23 @@
...
+PROTOCOL_ID = b"notdv5"
```
carver, line 47: Brainstorm:
- "ethdat"
- "estate"
- ~"ethp2p"~
- "datnet"

tests/core/v5_1/alexandria/test_content_storage.py
```diff
@@ -419,78 +402,66 @@ def test_content_storage_atomic_batch_fuzzy(data):
...
-        # `base_storage` **after** the batch operations.
-        storage_a = MemoryContentStorage()
+    # storage_a is managed such that it should be equivalent to
+    # `base_storage` **after** the batch operations.
+    storage_a = ContentStorage.memory()
 
-        # storage_a is managed such that it should be equivalent to
-        # `base_storage` **during** the batch operations.
-        storage_b = MemoryContentStorage()
+    # storage_a is managed such that it should be equivalent to
```
carver, line 416:
```suggestion
    # storage_b is managed such that it should be equivalent to
```
?

carver (commented):
Ignored all the deleted stuff, and any new code that seems to call methods outside the PR.

Other than that seems reasonable. :+1: Plus who doesn't love 10k lines of deletion :stuck_out_tongue:

ddht/v5_1/constants.py
pipermerriam: this is a leftover from guilherme's edits, we can leave it as `discv5`.  Nothing wrong with being part of the main network.

tests/core/v5_1/alexandria/test_alexandria_network.py
```diff
@@ -150,3 +151,101 @@ async def test_network_explore(tester, alice):
...
+                    content=response_content,
+                    request_id=request.request_id,
+                )
+
+            nursery.start_soon(_respond)
+
+            with trio.fail_after(2):
+                response_a = await alice_alexandria_network.find_content(
+                    bob.node_id,
+                    content_key=b'test-key',
```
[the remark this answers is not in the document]
pipermerriam: > Also, we have to decide what to do when someone sends us bad data. Do we simply ignore it? Drop the peer? Stop asking them for anything over the next n seconds?

For what we're planning to use this codebase for, I don't think this will matter that much since we should be operating in a non-adversarial environment for a while, and once we're not, the python client isn't likely to see continued development.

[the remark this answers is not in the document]
carver: Right, more thinking about the spec & recommendation for other clients. Agree it's not a blocker for this PR :+1:

ddht/v5_1/alexandria/rpc_handlers.py
```diff
@@ -102,58 +98,39 @@ async def do_call(self, params: ContentKey) -> str:
...
-        "alexandria_deletePinnedContent",
-        DeleteContentHandler(network.pinned_content_storage),
+        "alexandria_getContent",
+        GetContentHandler(network.storage),
     )
     yield (
-        "alexandria_deleteCommonsContent",
-        DeleteContentHandler(network.commons_content_storage),
+        "alexandria_deleteContent",
+        DeleteContentHandler(network.storage),
```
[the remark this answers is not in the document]
pipermerriam: I'm assuming you mean the JSON-RPC endpoint...

If yes... I just made the endpoints that seemed necessary to manage the database externally...

pipermerriam:
This is no longer actually *alexandria*, but that doesn't matter too much right now....  We can pretty easily rename if it seems appropriate.

PR #342, opened 2021-03-26
Replace encoded message cipher with empty bytes for WHOAREYOU packet

njgheorghita:
## What was wrong?
Packets using `EmptyMessage` for the message data, when the packet is `aesgcm_encrypt`'ed would result with a populated `message_cipher_text` field rather than simply an empty bytestring. 

Summary of approach.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/ddht/v5_1/packets.py
+++ b/ddht/v5_1/packets.py
@@ -10,7 +10,7 @@ from eth_typing import NodeID
 from eth_utils.toolz import take
 import rlp
 
-from ddht.base_message import BaseMessage
+from ddht.base_message import BaseMessage, EmptyMessage
 from ddht.constants import UINT8_TO_BYTES
 from ddht.encryption import aesctr_decrypt_stream, aesctr_encrypt, aesgcm_encrypt
 from ddht.exceptions import DecodingError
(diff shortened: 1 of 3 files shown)


ddht/v5_1/packets.py
```diff
@@ -244,12 +244,17 @@ def prepare(
...
         authenticated_data = b"".join((iv, header.to_wire_bytes(), auth_data_bytes,))
-        message_cipher_text = aesgcm_encrypt(
-            key=initiator_key,
-            nonce=aes_gcm_nonce,
-            plain_text=message.to_bytes(),
-            authenticated_data=authenticated_data,
-        )
+        # if empty, don't use an encrypted empty bytestring as the message_cipher_text field
+        if message.to_bytes():
+            message_cipher_text = aesgcm_encrypt(
```
pipermerriam, line 249: *nitpick*

I would restructure this in one of two ways.

```python
# This one simply avoids having to encode the message twice...
message_plain_text = message.to_bytes()

if not message_plain_text:
    message_cypher_text = b''
else:
    message_cypher_text = aesgcm_encrypt(...)
```

But I think the above is actually wrong, because in theory, it is legitimate for a message to be empty (but not actually be `EmptyMessage`... and the above over-matches, which on the receiving end would mean that I *think* the decryption would fail.

```python
if type(message) is EmptyMessage:
    message_cypher_text = b''
else:
    message_cypher_text = aesgcm_encrypt(...)
```

This doesn't overmatch and still accomplishes the same thing...  I think I like the later, though I think either will work.

pipermerriam (approved):
good to merge however you want to address this.

PR #344, opened 2021-05-12
Drop old sessions

carver:
## What was wrong?

Old sessions never got dropped. Also, when binding to loopback addresses, ddht would crash when trying to send to non-local addresses.

## How was it fixed?

Track session activity, and time out if peer doesn't send any messages for a while.

Explicitly warn the user about the localhost issue, instead of failing spectacularly later.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/ddht/v5_1/abc.py
+++ b/ddht/v5_1/abc.py
@@ -4,6 +4,7 @@ from typing import (
     Any,
     AsyncContextManager,
     Collection,
+    Container,
     NamedTuple,
     Optional,
     Protocol,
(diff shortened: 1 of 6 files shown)


ddht/v5_1/abc.py
```diff
@@ -207,6 +228,10 @@ def initiate_session(
     def receive_session(self, remote_endpoint: Endpoint) -> SessionAPI:
         ...
 
+    @abstractmethod
+    def __contains__(self, session_id: uuid.UUID) -> bool:
```
pipermerriam, line 232: *nitpick*

Changing this class to inherit from `collections.abc.Container` will do this automatically.

pipermerriam (approved):

PR #347, opened 2021-06-03
Source nodes from content logarithmic distance instead of actual distance

kdeme:
## What was wrong?

Receiving a FindContent request would error out at https://github.com/ethereum/ddht/blob/12216c70e1411adf53f318b42800e9f1b2b11703/ddht/v5_1/alexandria/network.py#L318
This is because the regular distance is being passed: https://github.com/ethereum/ddht/blob/12216c70e1411adf53f318b42800e9f1b2b11703/ddht/v5_1/alexandria/network.py#L379
Noticed this issue when testing interopt with Nim implementation.

## How was it fixed?

Pass the logarithmic distance instead.  Calculation added similar like it is already being done here https://github.com/ethereum/ddht/blob/743f3f3eaabb3b864b70e1bf78d441b5f0bdad46/ddht/kademlia.py#L108

#### Cute Animal Picture
image


carver (approved):
LGTM!

